PR #189, opened 2014-12-12
Start work on primitive, caseclass restrictions

ianoc:
Still a WIP:

Needs a test or two.

But it works, it only passes on case classes whose fields are all primitives. 

Use case:

Scalding JDBC

Aiming for ultimate usage to be along the lines of

`MysqlJdbcSource[MyType](connectionProps) extends TypedSink[MyType]`

So auto generate how to do whatever column mappings needed from the type, so restrictions on its shape are probably a reasonable first stab.


isnotinvain:
@johnynek @ianoc I'm working on doing a release, would you like this included?

ianoc:
No, you can leave this out of the release, not fully baked yet

ianoc:
Not a big need of this, we've done lots of macros since. Killing

PR #193, opened 2015-01-10
FIX: gzip Bijection resource leak.

plaflamme:
GZIPOutputStream#finish() does not relinquish all resources.

Specifically the Deflater instance used internally will release resources when end() called which only happens on close() or when finalize() is invoked by the garbage collector.

Thus, the use of finish() without close() relies on finalizers to cleanup resources which is a Bad Thing(tm).

The same logic was applied to the reverse Bijection.


ianoc:
Looks great, thanks for the contribution

PR #195, opened 2015-02-14
Allow building bijection-scrooge for scala 2.11

colinmarc:
I'm assuming it was disabled because scrooge wasn't available until recently,
but I was able to build without any issues.


colinmarc:
Err, let me try that again.

PR #198, opened 2015-02-21
Make property tests work again

MansurAshraf:
Tests were not running before. This change ensure tests are running (but not necessarily passing ...yet)

### DIFF
--- a/bijection-avro/src/test/scala/com/twitter/bijection/avro/GenericAvroCodecLaws.scala
+++ b/bijection-avro/src/test/scala/com/twitter/bijection/avro/GenericAvroCodecLaws.scala
@@ -13,19 +13,15 @@ limitations under the License.
 */
 package com.twitter.bijection.avro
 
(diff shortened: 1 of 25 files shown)


johnynek:
Looks good to me, but question: how can we be sure the tests are all running now? Can we set scalatest to error if you have a property that never asserts?

MansurAshraf:
I inverted the assert conditions in BaseProperties and made sure that tests run and fail.

johnynek:
@MansurAshraf cool. Can you comment on what the PR in #197 should do to be compatible with this style?

johnynek:
The URL injection test failed. It is flakey. Can you switch it from using the default `Arbitrary.arbitrary[String]: Gen[String]` to something like:

```scala
Arbitrary.arbitrary[String].filter(s => Try(new Url(s)).isSuccess)
```
then we only start with strings that won't throw.

johnynek:
Thanks for doing this! Next stop: code coverage.

PR #202, opened 2015-03-05
JsonScalaCodec doesn't need Base64 encoding.

dln:
Speculative pull request here (This may be entirely wrong). :)

Is there good reason to Base64 encode the JSON output generated by TJSONProtocol?

### DIFF
--- a/bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
+++ b/bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
@@ -19,6 +19,7 @@ package com.twitter.bijection.scrooge
 import com.twitter.bijection.{ Bijection, Injection }
 import com.twitter.bijection.Inversion.attempt
 import com.twitter.scrooge._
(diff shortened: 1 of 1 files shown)


johnynek:
I'm confused where Base64 is happening in the code? Can you clarify?

dln:
Hmm, had another look in scrooge, and the default encoder in ThriftStructSerializer is Base64StringEncoder: https://github.com/twitter/scrooge/blob/master/scrooge-serializer/src/main/scala/com/twitter/scrooge/ThriftStructSerializer.scala#L12

It's used here: https://github.com/twitter/bijection/blob/develop/bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala#L50

But I guess we should just use the JsonThriftSerializer instead of ThriftStructSerializer, rather than my hack?

johnynek:
I'm happy to have the change to fix this. I didn't know it was base64 encoding. Can you add a test that exercises this somehow?

johnynek:
Happy to revisit this later, it looks like it got lost here.

It would be nice to have some tests that assert that output looks a certain way (or can be parsed by some standard json library or something).

PR #203, opened 2015-03-15
Add implicit conversion for GZippedBase64String to String

jbripley:
- Include unit test for both GZippedBase64String and Base64String to String injections


johnynek:
Thanks!

Issue #205, opened 2015-03-31
use wart remover


johnynek:
we should turn on wart-remover: https://github.com/puffnfresh/wartremover to disable some aspects of scala that lead to bad errors or unsafe code.

Inferring Any, any2addString, Unit coercion are a few that look like they should definitely be on. I'd like to do more, but we may have some performance reasons for not turning on all the safety checks.

PR #206, opened 2015-03-31
First pass at some wart remover stuff, fixing things it mentioned

ianoc:


bijection-core/src/main/scala/com/twitter/bijection/CollectionInjections.scala
```diff
@@ -81,16 +81,22 @@ trait CollectionInjections extends StringInjections {
       }
       override def invert(d: D): Try[C] = {
         val builder = dc()
-        d foreach { b =>
-          val thisB = inj.invert(b)
+        var failed = false
+        val iter = d.toIterator
+        while (iter.hasNext && !failed) {
```
ianoc: I made this do the iterator approach, but the real thing wart remover was complaining about was the use of return

johnynek:
addresses #205

johnynek:
this seems to have bitrot quite a bit. If you want to take another stab later Ian, you are always welcome! :)

PR #207, opened 2015-04-06
upgrade testing libraries

MansurAshraf:
upgrade scalacheck and sclalatest


johnynek:
I'm fine with this, but what is the story with harmonizing with the internal twitter versions?

@ianoc any concern?

ianoc:
No concerns from me, I believe this is Mansur doing that harmonization

ianoc:
Merging since green, and no one seems to have any issues

PR #208, opened 2015-04-21
add Buf <-> Array[Byte] bijection

dschobel:
This PR defines a bijection between Array[Byte] and com.twitter.io.Buf with a bonus version bump of the util-core lib.

### DIFF
--- a/bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
+++ b/bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
@@ -17,6 +17,7 @@ limitations under the License.
 package com.twitter.bijection.twitter_util
 
 import com.twitter.bijection.{ AbstractBijection, Bijection, ImplicitBijection }
+import com.twitter.io.Buf
 import com.twitter.util.{ Future => TwitterFuture, Try => TwitterTry, Promise => TwitterPromise, Return, Throw, FuturePool }
 
 import scala.concurrent.{ Future => ScalaFuture, Promise => ScalaPromise, ExecutionContext }
@@ -121,6 +122,12 @@ trait UtilBijections {
     new AbstractBijection[FuturePool, ExecutionContext] {
       override def apply(pool: FuturePool) = new TwitterExecutionContext(pool)
       override def invert(context: ExecutionContext) = new ScalaFuturePool(context)
+    }
+
+  implicit def byteArrayBufBijection: Bijection[Array[Byte], Buf] =
+    new AbstractBijection[Array[Byte], Buf] {
+      override def apply(bytes: Array[Byte]) = Buf.ByteArray.Owned(bytes)
+      override def invert(buf: Buf) = Buf.ByteArray.Owned.extract(buf)
     }
 }
 
--- a/bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
+++ b/bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
@@ -125,6 +125,10 @@ trait UtilBijections {
     }
 
   object Owned {
+    /**
+     * A bijection using Buf's Owned api which minimizes copying of the underlying
+     * array data but places the onus of immutability on the user.
+     */
     implicit def byteArrayBufBijection: Bijection[Array[Byte], Buf] =
       new AbstractBijection[Array[Byte], Buf] {
         override def apply(bytes: Array[Byte]) = Buf.ByteArray.Owned(bytes)
@@ -133,6 +137,10 @@ trait UtilBijections {
   }
 
   object Shared {
+    /**
+     * A bijection using Buf's Shared api which avoids sharing state at the cost
+     * added allocations for defensive copies.
+     */
     implicit def byteArrayBufBijection: Bijection[Array[Byte], Buf] =
       new AbstractBijection[Array[Byte], Buf] {
         override def apply(bytes: Array[Byte]) = Buf.ByteArray.Shared(bytes)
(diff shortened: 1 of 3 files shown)


bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -122,6 +123,12 @@ trait UtilBijections {
       override def apply(pool: FuturePool) = new TwitterExecutionContext(pool)
       override def invert(context: ExecutionContext) = new ScalaFuturePool(context)
     }
+
+  implicit def byteArrayBufBijection: Bijection[Array[Byte], Buf] =
+    new AbstractBijection[Array[Byte], Buf] {
+      override def apply(bytes: Array[Byte]) = Buf.ByteArray.Owned(bytes)
```
mosesn: idk should this be owned or shared?

project/Build.scala
```diff
@@ -226,7 +226,7 @@ object BijectionBuild extends Build {
 
   lazy val bijectionUtil = module("util").settings(
     osgiExportAll("com.twitter.bijection.twitter_util"),
-    libraryDependencies += "com.twitter" %% "util-core" % "6.20.0"
+    libraryDependencies += "com.twitter" %% "util-core" % "6.23.0"
```
ianoc: is the bump required?

dschobel: it's not a frivolous bump if that's what you mean, if we want the new Buf apis (Shared/Owned) we need it, but it's also possible to define a bijection for the 6.20 Bufs

ianoc: No thats ok, once its required to access the API's you need for the change the bump is fine

ianoc:
Looks fine to me, but best to resolve with Moses about Owned or Shared, I don't have enough context on Buf to know the best answer there.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
dschobel: The [ChannelBuffer bijection](https://github.com/twitter/bijection/blob/71ed7b50d3ef4b5083060457dec6e52b6f65bad8/bijection-netty/src/main/scala/com/twitter/bijection/netty/ChannelBufferBijection.scala#L34) wraps the byte array on application and copies on extraction so this jives with the wrapping half but not the extraction. 

I can see a case for defining both versions and letting folks import the semantics they prefer? otoh, if users care about defensive copies, they can fairly trivially compose that onto any use of this version.

I do think making the shared impl the default is a mistake since this code tends to be used for serialization in batch jobs and as the hyper-doop folks showed is frequently a performance bottleneck.

mosesn: Hmm, maybe we should namespace them slightly differently and supply both?  I feel like the point of Owned vs Shared was that we wanted folks to have to think about what it was OK to do with the data, so maybe we should do it here too (since we intend for this to be an extension of our API).

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -122,6 +123,22 @@ trait UtilBijections {
       override def apply(pool: FuturePool) = new TwitterExecutionContext(pool)
       override def invert(context: ExecutionContext) = new ScalaFuturePool(context)
     }
+
+  object Owned {
```
johnynek: can we get some comments here to explain the differences? I guess in one case (shared?) you are copying the underlying buffer, but in the other you are not, which only makes it safe if you "own" the Buf/Array[Byte] in question. Is that right?

dschobel: That's right, it describes ownership of the underlying byte array with implications of defensive copying. I'll add a note.

ianoc: we now have objects Owned and shared under twitter_util, the namespacing there doesn't make too much sense. A sub package or object to put them into a Buf namespace so it aligns more with util I think would be good here

mosesn: actually, could we add Shared / Owned type aliases, and then use different type classes for .as[Shared] vs .as[Owned] or is scala too smart for that?

ianoc: I don't believe that'll work anyway -- You could though add them in as AnyVal case classes here, and make the injections be to those. It would be a similar strategy as to how the Base64String stuff is done

johnynek: something like:

```scala
case class Owned(bytes: Array[Byte]) extends AnyVal
case class Shared(bytes: Array[Byte]) extends AnyVal
```
then you can do `buf.as[Shared].bytes` if we want to go that way. I kind of like usage of types for this kind of decoration.

mosesn: Those wouldn't work for APIs that need to take Bufs though, right?  You'd need to do `bytes.as[Owned].unwrap`, which is practically as bad as

```scala
val Buf.Owned(bytes) = buf
val buf = Buf.Owned(bytes)
```

I think @ianoc's idea for a Buf.Owned namespace is reasonable.

dschobel: @mosesn / @johnynek / @ianoc everybody happy with a com.twitter.bijection.twitter_util.Buf namespace to hold the Shared and Owned objects?

ianoc: Sure, good with me

mosesn: works for me

johnynek:
Thanks @dschobel ! we're going to get this out in the next few days.

PR #210, opened 2015-05-06
Try bumping various scoverage deps versions to see if works better

ianoc:


sritchie:
This is handled by codecov, now.

PR #212, opened 2015-05-11
Update travis & sbt scala versions to 2.10.5 and 2.11.5

jnievelt:
These should be the only non-5 references to 2.10 & 2.11


ianoc:
LGTM -- merge when green

PR #213, opened 2015-05-12
Fixup unidoc

ianoc:
So unidoc concats the classpaths when it does its own compile pass. This means the classpath doesn't use any ivy eviction or sane ordering.

Thus we wound up with an old version of util via the scrooge-serializer, and we were getting an older version of joda-time thats built into jruby-complete(worst dep ever), which hbase depends on.

2.10.5 docs still barf due to the compiler plugin for quasiquotes macros


johnynek:
lgtm

PR #215, opened 2015-05-16
Removes unnecessary HBase injections

benpence:
The injection roles are satisfied by their bijection counterparts. In fact, explicitly calling the injections can cause a `StackOverflowError` as the injections satisfy the bijection implicit dependencies and vice versa.

Thank you @ianoc for your help.


MansurAshraf:
How about adding some tests that use injections?

johnynek:
1) we should keep the Injections, not the Bijections (which were a hack we used before we added the Injection type).

2) If we don't want to change the code much, just don't make the first set of Injections that Mansur added, but the last set (using ImmutableBytesWritable), should still have been fine.

ianoc:
the last set were the ones causing the stack overflow errors ben saw

johnynek:
Well, in any case, we should keep Injections here, not the Bijection with the `@@` type. In fact, I think we should remove the `@@` type from the library. I've never seen anyone use it and it confuses new users.

benpence:
OK I'll work to remove the bijections and rewrite the tests

PR #217, opened 2015-05-28
Migrates hbase bijections to injections because they weren't actually bijections

benpence:
Also fixes bug where the entire underlying byte array of ImmutableBytesWritable
was being copied, although only a slice of it was valid data. Adds unit test to
confirm that it works.


benpence:
Please merge #218 before merging this one.

bijection-hbase/src/main/scala/com/twitter/bijection/hbase/HBaseInjections.scala
```diff
@@ -14,31 +14,103 @@
...
+  implicit lazy val bigDecimal2BytesWritableInj =
+    new ImmutableBytesWritableInjection[BigDecimal] {
+      override def invert(b: ImmutableBytesWritable) = Try {
+        Bytes.toBigDecimal(b.get, b.getOffset, b.getLength)
+      }
+    }
+  implicit lazy val bytes2BytesWritableInj: Injection[Array[Byte], ImmutableBytesWritable] =
+    new AbstractInjection[Array[Byte], ImmutableBytesWritable] {
+      override def apply(a: Array[Byte]) = new ImmutableBytesWritable(a)
+      override def invert(b: ImmutableBytesWritable) = Success(b.copyBytes)
```
MansurAshraf: should this be ```Try{b.copyBytes}```? Are we positive this can never throw an exception?

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsLaws.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.scalacheck.Arbitrary
+
+import com.twitter.bijection.{ CheckProperties, BaseProperties, Injection }
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable
+
+object HBaseInjectionsLaws {
+  private[this] val arbitraryByteArray = implicitly[Arbitrary[Array[Byte]]]
+
+  implicit val arbitaryImmutableBytesWritable: Arbitrary[ImmutableBytesWritable] =
+    Arbitrary { arbitraryByteArray.arbitrary.map(new ImmutableBytesWritable(_)) }
```
MansurAshraf: you can also do

```
Arbitrary.arbitrary[Array[Byte]].map(new ImmutableBytesWritable(_)))

```
then you dont need ```arbitraryByteArray``` on line 25

bijection-hbase/src/main/scala/com/twitter/bijection/hbase/HBaseInjections.scala
```diff
@@ -14,31 +14,103 @@
...
+    }
+
+  /**
+   * ImmutableBytesWritable injections avoid copying when possible and use the
+   * slice (offset and length) of the underlying byte array.
+   */
+  implicit lazy val string2BytesWritableInj =
+    new ImmutableBytesWritableInjection[String] {
+      override def invert(b: ImmutableBytesWritable) =
+        Option(Bytes.toString(b.get, b.getOffset, b.getLength))
```
MansurAshraf: Is the whole ```Option(...).toTry``` done to avoid a null check? IMO it would be more efficient to just do 

```
Try{
  if(null check) sys.error(...)
 else result
} 
```

bijection-hbase/src/main/scala/com/twitter/bijection/hbase/HBaseInjections.scala
```diff
@@ -14,31 +14,103 @@
...
-  implicit lazy val double2BytesWritableInj: Injection[Double, ImmutableBytesWritable] = fromBijectionRep[Double, ImmutableBytesWritable]
-  implicit lazy val float2BytesWritableInj: Injection[Float, ImmutableBytesWritable] = fromBijectionRep[Float, ImmutableBytesWritable]
-  implicit lazy val short2BytesWritableInj: Injection[Short, ImmutableBytesWritable] = fromBijectionRep[Short, ImmutableBytesWritable]
-  implicit lazy val boolean2BytesWritableInj: Injection[Boolean, ImmutableBytesWritable] = fromBijectionRep[Boolean, ImmutableBytesWritable]
-  implicit lazy val bigDecimal2BytesWritableInj: Injection[BigDecimal, ImmutableBytesWritable] = fromBijectionRep[BigDecimal, ImmutableBytesWritable]
+  // Array[Byte] injections
+  implicit lazy val string2BytesInj: Injection[String, Array[Byte]] =
+    buildCatchInvert(Bytes.toBytes: String => Array[Byte])(Bytes.toString)
+  implicit lazy val long2BytesInj: Injection[Long, Array[Byte]] =
+    buildCatchInvert(Bytes.toBytes: Long => Array[Byte])(Bytes.toLong)
```
MansurAshraf: I think ```buildCatchInvert(Bytes.toBytes)(Bytes.toLong)``` should be fine and Scala should infer the types?

benpence: Unfortunately, scala's type inference will not. However, there may be a less verbose type annotation.

ianoc: buildCatchInvert[Long,Array[Byte])(Bytes.toBytes)(Bytes.toLong) might work here?

bijection-hbase/src/main/scala/com/twitter/bijection/hbase/HBaseInjections.scala
```diff
@@ -14,31 +14,105 @@
...
+  implicit lazy val float2BytesInj: Injection[Float, Array[Byte]] =
+    buildCatchInvert(Bytes.toBytes: Float => Array[Byte])(Bytes.toFloat)
+  implicit lazy val short2BytesInj: Injection[Short, Array[Byte]] =
+    buildCatchInvert(Bytes.toBytes: Short => Array[Byte])(Bytes.toShort)
+  implicit lazy val double2BytesInj: Injection[Double, Array[Byte]] =
+    buildCatchInvert(Bytes.toBytes: Double => Array[Byte])(Bytes.toDouble)
+  implicit lazy val bigDecimal2BytesInj: Injection[BigDecimal, Array[Byte]] =
+    new AbstractInjection[BigDecimal, Array[Byte]] {
+      override def apply(a: BigDecimal) = Bytes.toBytes(a.underlying)
+      override def invert(b: Array[Byte]) = Try { Bytes.toBigDecimal(b) } map (BigDecimal(_))
```
ianoc: just use ()'s for the simple statement, and always . before map's

no ws after the map

bijection-hbase/src/main/scala/com/twitter/bijection/hbase/HBaseInjections.scala
benpence: So it does. The Wise Mr. Ian!

MansurAshraf:
@benpence Looks good. One last thing would be to match the Hbase version number we are using internally and see all tests still pass

bijection-hbase/src/main/scala/com/twitter/bijection/hbase/HBaseInjections.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * ImmutableBytesWritable injections avoid copying when possible and use the
+   * slice (offset and length) of the underlying byte array.
+   */
+  implicit lazy val string2BytesWritableInj =
+    new ImmutableBytesWritableInjection[String] {
+      override def invert(b: ImmutableBytesWritable) =
+        Try {
+          val str = Bytes.toString(b.get, b.getOffset, b.getLength)
```
johnynek: seems like it is better to do:
```scala
Bytes.toString(b.get, b.getOffset, b.getLength) match {
  case null => Failure(new Exception(s"Could not read string from: $b"))
  case str => Success(str)
}
```
no need to throw and catch or use `Try { }`

johnynek: you could use fast attempt:
https://github.com/twitter/bijection/blob/develop/bijection-macros/src/main/scala/com/twitter/bijection/macros/Macros.scala#L21

which is a macro and has no perf cost:
```scala
fastAttempt(b)(Bytes.toString(b.get, b.getOffset, b.getLength) match {
  case null => sys.error("Could not read string from: $b")
  case str => str
})
```

bijection-hbase/src/main/scala/com/twitter/bijection/hbase/HBaseInjections.scala
```diff
@@ -0,0 +1,118 @@
...
+          if (str == null) sys.error("Invalid string")
+          else str
+        }
+    }
+  implicit lazy val int2BytesWritableInj =
+    new ImmutableBytesWritableInjection[Int] {
+      override def invert(b: ImmutableBytesWritable) = Try {
```
johnynek: could use `fastAttempt(b)(Bytes.toInt(b.get, b.getOffset, b.getLength))`

and so on below.

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsLaws.scala
```diff
@@ -0,0 +1,104 @@
...
+import com.twitter.bijection.{ CheckProperties, BaseProperties, Injection }
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable
+
+object HBaseInjectionsLaws {
+  implicit val arbitaryImmutableBytesWritable: Arbitrary[ImmutableBytesWritable] =
+    Arbitrary(for {
+      arbByteArray <- Arbitrary.arbitrary[Array[Byte]]
+    } yield new ImmutableBytesWritable(arbByteArray))
+
+  implicit val equivImmutableBytesWritable = new Equiv[ImmutableBytesWritable] {
```
johnynek: this is not needed. This is the default.

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsSpecifications.scala
```diff
@@ -0,0 +1,161 @@
...
+      val sharedLongIBW = new ImmutableBytesWritable(allBytes, 0, 8)
+      val sharedDoubleIBW = new ImmutableBytesWritable(allBytes, 8, 8)
+      val sharedIntIBW = new ImmutableBytesWritable(allBytes, 16, 4)
+
+      assert(Success(long) == long2BytesWritableInj.invert(sharedLongIBW))
+      assert(Success(double) == double2BytesWritableInj.invert(sharedDoubleIBW))
+      assert(Success(int) == int2BytesWritableInj.invert(sharedIntIBW))
+    }
+
+    "round trip String -> Array[Byte]" in {
```
johnynek: what are these testing? I don't think they are testing the ImmutableBytesWritable stuff, are they?

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsSpecifications.scala
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    "round trip Boolean -> Array[Byte]" in {
+      val expected = true
+      val bytes = Injection[Boolean, Array[Byte]](expected)
+      val result = Injection.invert[Boolean, Array[Byte]](bytes)
+      assert(Success(expected) == result)
+    }
+
+    "round trip String -> ImmutableBytesWritable" in {
```
johnynek: this is testing your code, but the stuff above is just retesting existing stuff in bijection, right?

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsLaws.scala
```diff
@@ -0,0 +1,104 @@
...
+
+import org.scalatest.{ PropSpec, MustMatchers }
+import org.scalatest.prop.PropertyChecks
+import org.scalacheck.Arbitrary
+
+import com.twitter.bijection.{ CheckProperties, BaseProperties, Injection }
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable
+
+object HBaseInjectionsLaws {
+  implicit val arbitaryImmutableBytesWritable: Arbitrary[ImmutableBytesWritable] =
```
johnynek: this is not robust enough. I think we need something like this:
```scala
for {
  bytes <- Arbitrary.arbitrary[Array[Byte]]
  offset <- Gen.choose(0, bytes.length - 1)
  len <- Gen.choose(0, bytes.length - offset)
} yield new ImmutableBytesWritable(bytes, offset, len)
```
so we can exercise having something that has offset/length other than the trivial values.

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsLaws.scala
```diff
@@ -0,0 +1,104 @@
...
+ * @author Muhammad Ashraf
+ * @since 7/10/13
+ */
+class HBaseInjectionsLaws extends CheckProperties with BaseProperties {
+  import HBaseInjections._
+
+  import HBaseInjectionsLaws.arbitaryImmutableBytesWritable
+  import HBaseInjectionsLaws.equivImmutableBytesWritable
+
+  property("String -> Array[Byte]") {
```
johnynek: the Array[Byte] ones are not from this code, are they? Can't they be removed?

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsLaws.scala
```diff
@@ -0,0 +1,104 @@
...
+  property("Boolean -> Array[Byte]") {
+    isLooseInjection[Boolean, Array[Byte]]
+  }
+
+  property("Short -> Array[Byte]") {
+    isLooseInjection[Short, Array[Byte]]
+  }
+
+  property("String -> ImmutableBytesWritable") {
+    isLooseInjection[String, ImmutableBytesWritable]
```
johnynek: does this have to be a LooseInjection? Can any of them be the standard approach (Int, Long, etc?) It is better to have the tests be as strict as can be.

I would think that some of them will pass with `isInjection`.

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsLaws.scala
benpence: You're talking about the `Equiv` impl? I added that earlier to ensure the test methods were picking up the right ones. I will remove it.

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsSpecifications.scala
benpence: This test ensures that, in a situation where 3 different values are backed by different parts of the same byte array, the hbase injection inversions still work.

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsSpecifications.scala
benpence: Most of the `T -> ImmutableBytesWritable` injections do not use the existing `T -> Array[Byte]` injections and, instead, use overloaded `Bytes` methods to reduce copying of the underlying byte array.

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsLaws.scala
benpence: OK, will do

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsLaws.scala
benpence: These `Array[Byte]` injections may be superfluous. Their purpose is for the situation where you have a value serialized as `Array[Byte]` with hbase encoding. I would need to know more about how people use hbase before I'm confident in removing them.

Perhaps you're speaking to the lack of the `Rep` type? I guess it's possible that these injections may enable users to make mistakes by growing the domain from `Array[Byte] @@ Rep[T]` to `Array[Byte]`.

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsLaws.scala
benpence: Yes you are correct. I just checked: the `ImmutableBytesWritable` `.equals` method *is* offset and length; therefore, the strict injection check should work with the global catchall `Equiv[T]`.

bijection-hbase/src/test/scala/com/twitter/bijection/hbase/HBaseInjectionsSpecifications.scala
johnynek: Oh I see. The HBaseInjections are available in both `Array[Byte]` and `ImmutableBytesWritable` form. got it.

johnynek:
Let's go ahead and merge. Looks like you didn't want to do my suggestions of using the macros to avoid the closures needed to use the `Try { }` syntax. That's fine.

PR #220, opened 2015-06-02
Simplifies hbase injections with fastAttempt macro

benpence:


bijection-hbase/src/main/scala/com/twitter/bijection/hbase/HBaseInjections.scala
```diff
@@ -58,53 +59,45 @@ object HBaseInjections {
   implicit lazy val string2BytesWritableInj =
     new ImmutableBytesWritableInjection[String] {
       override def invert(b: ImmutableBytesWritable) =
-        Try {
-          val str = Bytes.toString(b.get, b.getOffset, b.getLength)
-          if (str == null) sys.error("Invalid string")
-          else str
-        }
+        fastAttempt(b)(Bytes.toString(b.get, b.getOffset, b.getLength) match {
+          case null => sys.error("Could not read string from: $b")
```
johnynek: you need `s"Could not read ...` not the `s` to do interpolation. This is a common bug and a bit of a pain with interpolation.

I'd recommend actually: `s"$b decoded to null, which is disallowed."`

project/Build.scala
```diff
@@ -262,6 +262,7 @@ object BijectionBuild extends Build {
       "org.apache.hadoop" % "hadoop-core" % "1.0.4" % "provided->default"
     )
   ).dependsOn(bijectionCore % "test->test;compile->compile")
+   .dependsOn(bijectionMacros % "test->test;compile->compile")
```
johnynek: I think this is just `.dependsOn(bijectionMacros)` you don't need to depend on the test code (as we are doing in these other examples).

PR #221, opened 2015-06-03
Specialize the TBinaryProtocol read path up to ~2x speedups

ianoc:
bijectiondeserwin

Can see the prior delta in the posted image. The macro stays below but the others that do full deserialization converge with this change.

### DIFF
--- a/bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
+++ b/bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
@@ -18,6 +18,7 @@ package com.twitter.bijection.scrooge
 
 import com.twitter.bijection.{ Bijection, Injection }
 import com.twitter.bijection.Inversion.attempt
+import com.twitter.bijection.macros.Macros
 import com.twitter.scrooge._
 import org.apache.thrift.protocol.TJSONProtocol
 
@@ -44,7 +45,7 @@ class BinaryScalaCodec[T <: ThriftStruct](c: ThriftStructCodec[T])
   }
 
   override def apply(item: T) = thriftStructSerializer.toBytes(item)
-  override def invert(bytes: Array[Byte]) = attempt(bytes){ bytes =>
+  override def invert(bytes: Array[Byte]) = Macros.fastAttempt(bytes){
     c.decode(TArrayBinaryProtocol(TArrayByteTransport(bytes)))
   }
 }
--- a/bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
+++ b/bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
@@ -35,10 +35,19 @@ object BinaryScalaCodec {
 }
 
 class BinaryScalaCodec[T <: ThriftStruct](c: ThriftStructCodec[T])
-  extends ScalaCodec(new ThriftStructSerializer[T] {
+  extends Injection[T, Array[Byte]] {
+  import com.twitter.bijection.thrift.{ TArrayByteTransport, TArrayBinaryProtocol }
+
+  lazy val thriftStructSerializer = new ThriftStructSerializer[T] {
     override def codec = c
     val protocolFactory = new TBinaryProtocol.Factory
-  })
+  }
+
+  override def apply(item: T) = thriftStructSerializer.toBytes(item)
+  override def invert(bytes: Array[Byte]) = attempt(bytes){ bytes =>
+    c.decode(TArrayBinaryProtocol(TArrayByteTransport(bytes)))
+  }
+}
 
 object CompactScalaCodec {
   def apply[T <: ThriftStruct](c: ThriftStructCodec[T]) =
--- a/bijection-thrift/src/main/scala/com/twitter/bijection/thrift/TArrayBinaryProtocol.scala
+++ b/bijection-thrift/src/main/scala/com/twitter/bijection/thrift/TArrayBinaryProtocol.scala
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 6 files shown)


bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
```diff
@@ -35,10 +35,19 @@ object BinaryScalaCodec {
...
+  import com.twitter.bijection.thrift.{ TArrayByteTransport, TArrayBinaryProtocol }
+
+  lazy val thriftStructSerializer = new ThriftStructSerializer[T] {
     override def codec = c
     val protocolFactory = new TBinaryProtocol.Factory
-  })
+  }
+
+  override def apply(item: T) = thriftStructSerializer.toBytes(item)
+  override def invert(bytes: Array[Byte]) = attempt(bytes){ bytes =>
```
johnynek: what about Macros.fastAttempt here to avoid the closure. :) Maybe no win, but it can't hurt.

bijection-thrift/src/main/scala/com/twitter/bijection/thrift/ThriftCodecs.scala
```diff
@@ -63,7 +63,23 @@ object BinaryThriftCodec {
...
+
+  protected lazy val prototype = klass.newInstance
+
+  override def apply(item: T) = {
+    val baos = new ByteArrayOutputStream
+    item.write(factory.getProtocol(new TIOStreamTransport(baos)))
+    baos.toByteArray
+  }
+  override def invert(bytes: Array[Byte]) = attempt(bytes) { bytes =>
+    val obj = prototype.deepCopy
```
johnynek: I wonder if it is better to do `.newInstance` here rather than deepCopy?

bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
```diff
@@ -35,10 +35,19 @@ object BinaryScalaCodec {
 }
 
 class BinaryScalaCodec[T <: ThriftStruct](c: ThriftStructCodec[T])
```
johnynek: shouldn't we do the same trick on CompactScalaCodec too?

bijection-thrift/src/main/scala/com/twitter/bijection/thrift/ThriftCodecs.scala
johnynek: also, maybe Macros.fastAttempt

bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
ianoc: Perf sensitive code, might as well take every little bit.

bijection-thrift/src/main/scala/com/twitter/bijection/thrift/ThriftCodecs.scala
ianoc: So thats a good question, not sure why its always deep copy. I took it from the existing code. Any idea if there was a bug/functional reason between the two?

bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
ianoc: Probably, it'll take more work to go through its serialization to ensure its sane though. Do we know if its used much to make it worth it?

mosesn:
This isn't feasible for dependency reasons, but it would be nice if we didn't duplicate work.  We have an implementation of TBinaryProtocol in finagle to minimize allocations:

https://github.com/twitter/finagle/blob/master/finagle-thrift/src/main/scala/com/twitter/finagle/thrift/Protocols.scala#L51

It would be neat to be able to use your optimizations when possible too.

ianoc:
Yeah alas not obvious for that, a scrooge-thrift type package with slim deps might be a good middle ground to use for everything

bijection-scrooge/src/main/scala/com/twitter/bijection/scrooge/ScroogeCodecs.scala
johnynek: Somehow I missed the new protocol implementation (was that left out of the first commit?) I get it now.

ianoc: Yeah sorry I did a got commit -am and missed the files. :/

johnynek:
looks good. Nice work.

bijection-thrift/src/main/scala/com/twitter/bijection/thrift/TArrayBinaryProtocol.scala
```diff
@@ -0,0 +1,158 @@
...
+      ((transport.buf(off + 7) & 0xffL))
+  }
+
+  def readDouble: Double =
+    java.lang.Double.longBitsToDouble(readI64)
+
+  def readString: String =
+    try {
+      val size = readI32
+      val s = new String(transport.buf, transport.bufferPos, size, "UTF-8")
```
isnotinvain: you could try the trick used here:
https://github.com/apache/parquet-mr/blob/d6f082b9be5d507ff60c6bc83a179cc44015ab97/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java#L90

Charset.decode instead of new String

ianoc: It doesn't mention any relative performance numbers, much faster?

isnotinvain: If this is the one that I'm thinking of, up to 40x faster, but it only applies to java6 and was fixed in java7. But I could be wrong.

ianoc: Well we are moving to java8, so not sure it matters much if fixed in java7

isnotinvain: I can't really tell, a read a blog post just now that said they fixed it for single-byte charsets, but not yet for multi-byte charsets like utf8.

PR #223, opened 2015-07-15
Add toBinaryWithSchema in order to support Avro generated Scala case classes

mariussoutier:
Add `toBinaryWithSchema` in order to support Avro generated Scala case classes.

Otherwise Scala classes are not recognized as specific records because they don't have a public field called `SCHEMA$`.

### DIFF
--- a/bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
+++ b/bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
@@ -109,6 +109,18 @@ object SpecificAvroCodecs {
   }
 
   /**
+   * Returns Injection capable of serializing and deserializing a compiled avro record using org.apache.avro.io.BinaryEncoder.
+   * Accepts a schema to support generated Scala case classes.
+   * @tparam T compiled Avro record
+   * @return Injection
+   */
+  def toBinaryWithSchema[T <: SpecificRecordBase: ClassTag](schema: Schema): Injection[T, Array[Byte]] = {
+    val writer = new SpecificDatumWriter[T](schema)
+    val reader = new SpecificDatumReader[T](schema)
+    new BinaryAvroCodec[T](writer, reader)
+  }
+
+  /**
    * Returns Injection capable of serializing and deserializing a generic avro record using org.apache.avro.io.JsonEncoder to a
    * UTF-8 String
    * @tparam T compiled Avro record
--- a/bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
+++ b/bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
@@ -110,12 +110,13 @@ object SpecificAvroCodecs {
 
   /**
    * Returns Injection capable of serializing and deserializing a compiled avro record using org.apache.avro.io.BinaryEncoder.
-   * Accepts a schema to support generated Scala case classes. This is not completely typesafe and will fail at runtime
-   * when you pass in the wrong schema.
+   * Fetches the schema from the specified class in order to be compatible with generated Scala classes.
    * @tparam T compiled Avro record
    * @return Injection
    */
-  def toBinaryWithSchema[T <: SpecificRecordBase](schema: Schema): Injection[T, Array[Byte]] = {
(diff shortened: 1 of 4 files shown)


bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
```diff
@@ -109,6 +109,18 @@ object SpecificAvroCodecs {
...
   /**
+   * Returns Injection capable of serializing and deserializing a compiled avro record using org.apache.avro.io.BinaryEncoder.
+   * Accepts a schema to support generated Scala case classes.
+   * @tparam T compiled Avro record
+   * @return Injection
+   */
+  def toBinaryWithSchema[T <: SpecificRecordBase: ClassTag](schema: Schema): Injection[T, Array[Byte]] = {
+    val writer = new SpecificDatumWriter[T](schema)
+    val reader = new SpecificDatumReader[T](schema)
+    new BinaryAvroCodec[T](writer, reader)
```
johnynek: does not look like this method would need `:ClassTag` to me. Am I missing something?

johnynek: This is also type-unsafe. If you pass the wrong schema here, I guess nothing will fail until you go to call apply on the injection. We should at least add a comment that it is the responsibility of the caller to verify that the Schema matches the type T. It is a shame that schemas are not typed.

mariussoutier: You are right about `ClassTag`, will remove it.

If the schema is similar (e.g. same record name) it seems to kind of work. When it's completely wrong, it throws `java.lang.ClassCastException: org.apache.avro.generic.GenericData$Record cannot be cast to avro.FiscalRecordScala`. If have added a test case and doc string.

mariussoutier: Too bad you can't just extract a schema from the class. I could instantiate it with the default constructor and call `getSchema`, but that sounds a bit hacky.

johnynek: Does that work? If all Avro classes have `.getSchema` and an empty constructor, I don't actually see that as hacky.

johnynek: But how would this be different than: https://github.com/twitter/bijection/pull/223/files#diff-5722c7a78a89c696ada627a7d0e51c7dR104

mariussoutier: `.getSchema` is part of `SpecificRecordBase` and the empty constructor required by the spec. The unit tests also run successfully.

The difference with `toBinary` is I'd pass in the schema to `SpecificDatum{Writer,Reader}`. If you don't, it will fail at runtime because Avro's `SpecificData` class can't find the SCHEMA$ field (Scala fields are always private). It's all about this stupid Java/Scala difference + reflection.

I've also opened a ticket with Avro directly, but I'm not sure they'll change this only because of Scala.

johnynek:
should we not just change how the existing `toBinary` works? How is that one working correctly now? If it will fail, we shouldn

mariussoutier:
The current `toBinary` works when the specific record class is recognized as such (i.e. a Java generated one with a public `SCHEMA$` field). Avro de/serialization is a runtime game :(

ianoc:
@johnynek any thoughts on this? with upcoming rls does it need some re-work/merge?

johnynek:
I just don't see why his new behavior should not be the default. Do you?

ianoc:
No, i don't know alot about avro i'd have to admit. But this change does look like maybe a strict improvement to a better API?

julianpeeters:
Yes, I think the new behavior should be the default. Avro has got to get a schema somehow, and preempting reflection by directly passing a schema not only opens the door to Scala records, but seems like better practice anyways? But my models aren't that large yet; is anybody going to be irritated by an extra [instantiation](https://github.com/mariussoutier/bijection/commit/b98cc7e9d483e9de1a62ee3c54e452218059ea5c#diff-5722c7a78a89c696ada627a7d0e51c7dR118) of every model?

mariussoutier:
Nice, thanks!

PR #224, opened 2015-07-20
En-threadsafe json injections for collections

bkirwi:
Hi!

Currently, the injections produced by the `JsonInjection` implicits are not threadsafe, since they reuse the same builder instance. If two different threads call into `invert` at the same time, both threads will clear and then add items to that builder, and the results of this are undefined.

This patch switches to using a brand-new builder instance for every call to invert. The resulting injections should be safe to use from multiple threads.

This involves a small change to the API: in particular, it drops the `fromBuilder` method and leans everything on `collectionJson`. Seemed mostly harmless, since the `fromCollection` method implements roughly the same functionality in a safer way -- but it *is* a breaking change.


johnynek:
Thanks for catching this!

bkirwi:
NP! Thanks for the pull.

PR #225, opened 2015-07-23
Upgrade sbt launcher script (sbt-extras)

isnotinvain:


isnotinvain:
Actually, this PR updates the sbt launcher, and tells travis to use it -- we don't have to do that, we're currently using whatever `sbt` is preconfigured in the travis environment, which I assume will be fixed soon: https://twitter.com/travisci/status/621055235038035968

johnynek:
looks like there is some escaping bug here. Legit failure.

isnotinvain:
what do you think about the previous comment though? this build wasn't using the ./sbt at all, it was using the pre-installed sbt

johnynek:
Well, the problem with using their installed sbt is that we are at their mercy, and we are not testing how we develop locally or how our contributors might use sbt. I think it is nicer to have the only dependency be a jvm, otherwise we will have to support a lot of sbt issues for new contributors.

So, I'd rather standardize on the launcher being in the state of the repo.

isnotinvain:
yeah sgtm, i'll try to fix this one

isnotinvain:
OK, think it's fixed. Looks like a property test found a bug though:

```
[info] - UUID -> String
[info] - URL -> String *** FAILED ***
[info]   IllegalArgumentException was thrown during property evaluation.
[info]     Message: invalid host wildcard specification
[info]     Occurred when passed generated values (
[info]       arg0 = http://鷕ﺌ遦ꆙ諪玄饆ᙆ䊡⺓壑著錮쾠䗚㧪ى᭲岻㒾픷挵壨頛㚚‛ᾍ䲊椄ᄀ뒾葊쟿旬䌡㝖俶펪佩昏᜞狖嘲僀ãꗷⶈ㢊ᗽ㫋*Ꞝ梞⬚츙.com
[info]     )
```

isnotinvain:
looks like it might be the *

johnynek:
We should fix that test. It does fail with randomly generated unicode, which is not a good way to generate URLs.

isnotinvain:
I've been trying (and failing) to find a string that repros this. Copying the above example doesn't seem to work (probably something is lost in the copy)

isnotinvain:
the odd thing is we do Try(new Url(_)), and the only other method called on this object is .toString. the error "invalid host wildcard specification" comes from here: https://github.com/openjdk-mirror/jdk/blob/jdk7/jdk7/master/src/share/classes/java/net/SocketPermission.java#L425 but I don't yet see why / how we end up there

isnotinvain:
seems like you get there by a) having a security manager and b) calling openConnection -- I don't know why we'd be doing b, and maybe a) has to do with travis setting one up?

isnotinvain:
turns out URL.java is pretty complicated :/

PR #228, opened 2015-09-03
Getting rid of a circular dependency between bijection-macros and chill

sriramkrishnan:
Bijection and Chill now have a circular dependency between each other. This is because  com.twitter.chill.Externalizer is used for the bijection-macros test.

I cherrypicked and simplified the Externalizer, and moved it into the TestHelpers in bijection-macros.

### DIFF
--- a/bijection-macros/src/test/scala/com/twitter/bijection/macros/MacroPropTests.scala
+++ b/bijection-macros/src/test/scala/com/twitter/bijection/macros/MacroPropTests.scala
@@ -8,7 +8,6 @@ import org.scalatest.prop.PropertyChecks
 
 import com.twitter.bijection._
 import com.twitter.bijection.macros._
-import com.twitter.chill.Externalizer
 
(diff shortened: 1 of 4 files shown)


johnynek:
Actually, I'm not so sure this is the best approach.

I think bijection-macros should be killed. We should move the macros into bijection (no need for a new package). Also, the IsCaseClass and MacroGenerated traits are really not useful. They were something generated when we were making the first few macros, but they are not valuable, as far as I can see (we later noticed other ways to do the same thing).

johnynek:
That said, we could merge this and just take on killing bijection-macros as an issue.

sriramkrishnan:
@johnynek yeah - my preference would be to kill bijection-macros as a separate issue. Thanks for looking at this.

johnynek:
added #229

sriramkrishnan:
That was quick - thanks!

PR #232, opened 2016-01-21
set the max file size to deal with #184

### LINKED ISSUE #184: Huge chain of anonymous classes leads to class file that is too big
This issue is seen here https://github.com/twitter/scalding/pull/1059

Digging around https://github.com/scala/pickling/issues/10 it looks like setting -Xmax-classfile-name may solve this, but this is an undesirable state for things to be in (if it can be avoided), especially since bijection is supposed to be a fairly dep-free, core lib that anyone can use.

johnynek:
closes #184

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -40,7 +40,10 @@ object BijectionBuild extends Build {
 
     parallelExecution in Test := true,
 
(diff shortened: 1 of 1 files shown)


ianoc:
lgtm, merge when green. Don't think we normally ever care about the class file name

PR #237, opened 2016-01-22
Make string Injections safer. Fix #199

### LINKED ISSUE #199: Injection.invert[String, Array[Byte]](Array[Byte](1, 2, 3, -1, -2, -127)) produces a string that cannot be inverted back
screen shot 2015-02-23 at 3 17 27 pm

johnynek:
close #199 

?r @ianoc 

Will you humor me on the atomics? :)

### DIFF
--- a/bijection-core/src/main/scala/com/twitter/bijection/StringInjections.scala
+++ b/bijection-core/src/main/scala/com/twitter/bijection/StringInjections.scala
@@ -18,9 +18,8 @@ package com.twitter.bijection
 
 import com.twitter.bijection.Inversion.attempt
 import java.net.{ URLDecoder, URLEncoder, URL }
-import java.nio.charset.{Charset, CodingErrorAction}
-import java.nio.ByteBuffer
-
+import java.nio.charset.{ Charset, CoderResult, CodingErrorAction }
+import java.nio.{ ByteBuffer, CharBuffer }
 import java.util.UUID
 import scala.annotation.tailrec
 import scala.collection.generic.CanBuildFrom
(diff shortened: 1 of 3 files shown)


bijection-core/src/main/scala/com/twitter/bijection/StringInjections.scala
```diff
@@ -32,9 +33,22 @@ trait StringInjections extends NumericInjections {
 
   def withEncoding(encoding: String): Injection[String, Array[Byte]] =
     new AbstractInjection[String, Array[Byte]] {
+      private[this] val decRef = {
+        val cs = Charset.forName(encoding)
+        new AtomicSharedState(() => cs.newDecoder.onUnmappableCharacter(CodingErrorAction.REPORT))
```
ianoc: is the charset threadsafe itself? 

Maybe we should have more like a () => {
val cs = Char...
})
? if threadsafe though doesn't seem like any reason to ofc

bijection-core/src/main/scala/com/twitter/bijection/StringInjections.scala
```diff
@@ -32,9 +33,22 @@ trait StringInjections extends NumericInjections {
...
+      }
+
       def apply(s: String) = s.getBytes(encoding)
       override def invert(b: Array[Byte]) =
-        attempt(b) { new String(_, encoding) }
+        // toString on ByteBuffer is nicer, so use it
+        attempt(ByteBuffer.wrap(b)) { bb =>
```
ianoc: We do an extra allocation of a ByteBuffer class here always now, should we maybe only put this in the error handling portion of things?

ianoc:
I've no problem with atomics :)

bijection-core/src/main/scala/com/twitter/bijection/StringInjections.scala
johnynek: the decoder requires a ByteBuffer, so bb is that. I can't seem to find a way to avoid that allocation.

ianoc: Oh ok, from reading the comment i thought we were using ByteBuffer so the toString in the event of an error was nicer. It looks like we mean the toString function on the decoder is better?

ianoc:
LGTM

bijection-core/src/main/scala/com/twitter/bijection/StringInjections.scala
johnynek: the attempt function includes a toSting of the argument, which in this case is the ByteBuffer.

PR #239, opened 2016-02-01
update the build

johnynek:


.gitignore
```diff
@@ -7,7 +7,7 @@
 #sbt
 /project/target/
 /project/project/target
-
+sonatype.sbt
```
ianoc: does one normally not put this into the ~/.sbt/0.13 folder ?

build.sbt
```diff
@@ -0,0 +1,307 @@
+import com.typesafe.tools.mima.plugin.MimaPlugin.mimaDefaultSettings
```
ianoc: Why move from project/Build.scala to a build.sbt? is this now the more blessed way?

project/plugins.sbt
```diff
@@ -1,12 +1,11 @@
...
 )
 
-addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.5.2")
-
-addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.6")
-
-addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.7.0")
-
-addSbtPlugin("com.typesafe.sbt" % "sbt-scalariform" % "1.3.0")
+addSbtPlugin("com.github.gseitz" % "sbt-release"     % "1.0.0")
```
ianoc: where does this formatting style come from ?

johnynek: me personally... :) I just cleaned it when adding do it.

ianoc: Could we use scalriform? the old one is closer to how we format most of our scala I suppose

build.sbt
johnynek: yes, and the plugins are a pain to config without it. There is some autoImport thing that is now standard, so people assume you are using that.

.gitignore
johnynek: well, sbt release has some possibly per project options, so I put it here, but it won't publish a dirty directory.

project/plugins.sbt
johnynek: Happy to. Can it run on .sbt files?

PR #243, opened 2016-02-16
Fix an integral float overflow for very large strings

johnynek:
we hit this with a 70MB string. Work around by removing a byte. Better to fix.

### DIFF
--- a/bijection-core/src/main/scala/com/twitter/bijection/StringInjections.scala
+++ b/bijection-core/src/main/scala/com/twitter/bijection/StringInjections.scala
@@ -53,7 +53,7 @@ trait StringInjections extends NumericInjections {
           val decBuf = decRef.get
           val dec = decBuf._1
           val buf = decBuf._2
-          val maxSpace = (b.length * dec.maxCharsPerByte).toInt + 1
+          val maxSpace = (b.length * (dec.maxCharsPerByte.toDouble)).toInt + 1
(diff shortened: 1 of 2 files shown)


bijection-core/src/main/scala/com/twitter/bijection/StringInjections.scala
```diff
@@ -53,7 +53,7 @@ trait StringInjections extends NumericInjections {
           val decBuf = decRef.get
           val dec = decBuf._1
           val buf = decBuf._2
-          val maxSpace = (b.length * dec.maxCharsPerByte).toInt + 1
+          val maxSpace = (b.length * (dec.maxCharsPerByte.toDouble)).toInt + 1
```
jnievelt: I'd put a comment, that for cases where it's possible to have 2B, just try using 2B.  I'd also drop the () and maybe just toDouble on the length, since that's the thing that can be large.

johnynek: actually, the bug is that dec.maxCharsPerByte returns a `Float`. That can only store all integers up to `2^{24}`. After that it steps up by 2. So the toInt truncates by 2 in that case, not rounding down by one.

I don't know what `2B` means in this case. Can you explain?

jnievelt: 2B = 2 billion (1 << 31), since that's how toInt handles larger values.

But yeah, Float -> Double makes more sense than integral -> Double, so that seems fine.

PR #245, opened 2016-03-20
upgrade Joda libraries:  joda-time 2.9.2, joda-convert 1.8

sullis:


johnynek:
Are there any binary compatibility issues that this resolves? We prefer to use the oldest version that is still binary compatible with the most recent so as not to force users to upgrade.

Really, joda should be a provided dependency, because all the consumers should declare which version of joda they are using in their project.

build.sbt
```diff
@@ -282,8 +282,8 @@ lazy val bijectionHbase = module("hbase").settings(
 lazy val bijectionJodaTime = module("jodatime").settings(
   osgiExportAll("com.twitter.bijection.jodatime"),
   libraryDependencies ++= Seq(
-    "joda-time" % "joda-time" % "2.3",
-    "org.joda" % "joda-convert" % "1.6"
+    "joda-time" % "joda-time" % "2.9.2",
```
johnynek, line 285: can we add `% "provided"` to this and the next line?

sritchie:
Closing - please open if this is still an issue and we can resume!

PR #247, opened 2016-05-03
JavaFuture <=> TwitterFuture bijection

BenFradet:
This implements a bijection between `JavaFuture`s and `TwitterFuture`s.

It originally came from [the storehaus-kafka module](https://github.com/twitter/storehaus/pull/286) because the kafka api uses `JavaFuture`.

The idea is to make it available here since it'd fit nicely with the other bijection between `ScalaFuture`s and `TwitterFuture`s.

credit to @johnynek for the implementation


bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -78,6 +80,32 @@ trait UtilBijections {
     }
   }
 
+  private val futureConverter = {
+    val converter = new JavaFutureToTwitterFutureConverter()
+    converter.start()
+    converter
+  }
```
BenFradet, line 96: If you have ideas on how to close the converter properly it'll be most welcome.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -78,6 +80,32 @@ trait UtilBijections {
...
+
+  /**
+   * Injection from java futures to twitter futures
+   */
+  implicit def twitter2JavaFuture[A]: Bijection[TwitterFuture[A], JavaFuture[A]] = {
```
johnynek: what if we take an `implicit JavaFutureToTwitterFutureConverter` here as a parameter, and then it is the responsibility of the user to create and eventually stop their futureConverter?

BenFradet:
oops `CompletableFuture` is java8 onwards

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
BenFradet: We could do that but wouldn't that complicate the usage of the api?

johnynek: yes, but having a thread started as a private val is kind of a non-starter, I think. It will impact serialization and it will consume resources for everyone who use this or not.

BenFradet: Yup :(

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/JavaFutureToTwitterFutureConverter.scala
```diff
@@ -0,0 +1,107 @@
...
+
+import scala.annotation.tailrec
+
+/**
+ * Utility class for converting Java futures to Twitter's
+ *
+ * @param waitTimeMs Time spent sleeping by the thread converting java futures to
+ *                   twitter futures when there are no futures to convert in ms
+ */
+private[twitter_util] class JavaFutureToTwitterFutureConverter(waitTimeMs: Long = 1000L) {
```
mosesn: This seems overly complicated (and induces a lot of latency).  How about instead the default is that we use a thread per future? the base implementation would be something like:

``` scala
import com.twitter.util.{FuturePool, Promise, Future}
import java.util.concurrent.{Future => JFuture}
import scala.util.control.NonFatal

class FuturePoolConverter(mayInterruptWhileRunning: Boolean) {
  def apply[A](jfuture: JFuture[A]): Future[A] = {
    val f = FuturePool.unboundedPool { jfuture.get() }
    val p = Promise.detached(f)
    p.setInterruptHandler { case NonFatal(e) =>
      if (p.detach()) {
        f.raise(e)
        jfuture.cancel(mayInterruptWhileRunning)
      }
    }
    p
  }
}
```

A more sophisticated converter could use a Timer explicitly, and do something similar to what the current implementation does.

``` scala
import com.twitter.util.{FuturePool, Promise, Timer, Try, Throw, Future}
import java.util.concurrent.{Future => JFuture}
import scala.util.control.NonFatal

class FuturePoolConverter(timer: Timer, checkFrequency: Duration, mayInterruptWhileRunning: Boolean) {
  def apply[A](jfuture: JFuture[A]): Future[A] = {
    val p = Promise[A]()
    val task = timer.schedule(checkFrequency) { 
      if (jfuture.isDone) {
        p.updateIfEmpty(Try(jfuture.get()))
        task.cancel()
      }
    }
    p.setInterruptHandler { case NonFatal(e) =>
      task.cancel()
      p.updateIfEmpty(Throw(e))
    }
    p
  }
}
```

Thoughts?

BenFradet: Wouldn't using one thread per future be a bit wasteful compared to the current approach?
On the other hand, this approach would simplify the API as there wouldn't be a need for an implicit parameter.  
Also, could you explain what you mean by "induces a lot of latency" compared to your second approach?

mosesn: @BenFradet I think we should still use the implicit parameter.

Yes, it would be wasteful.  That's why I also proposed the Timer approach, which I think is quite similar to your approach, and is appropriate for cases where there may be many futures.  The overhead will be mostly in memory for threads (if you have many concurrent futures you're waiting for) and in context switches (if you have many futures finishing at the same time).

A timer approach is almost always going to be relatively high latency, because they all work using tick mechanisms.  When you block, you move the work to the scheduler, which has the ability to work up a specific thread when there's work to do.  This means that a thread can be notified immediately if we're blocking, but if we're waiting for the next tick to check, it will add the latency until the next tick necessarily.  This will move the latency of discovering a future has finished from ~microseconds to ~milliseconds.

johnynek: thread per future seems like a bad idea to me.

I'm not crazy about the polling all the time, but maybe the current approach is equivalent to the timer anyway (assuming a well implemented `Timer`).

We could do both approaches:

``` scala
abstract class JavaFutureConverter {
  def apply[T](j: JFuture[T]): Future[T]
}

class FuturePoolJConverter(implicit fp: FuturePool) {
  ...
}

class TimerJConverter(t: Timer, d: ...) {

}
```

Then the `bijection` takes an implicit JavaFutureConverter which in some cases might be FuturePool based, in others timer/polling?

mosesn: @johnynek yep, that's exactly what I'm proposing.

BenFradet: @mosesn thanks for the explanation!

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -79,6 +81,22 @@ trait UtilBijections {
...
 
   /**
+   * Injection from java futures to twitter futures
+   */
+  implicit def twitter2JavaFuture[A](
+    implicit converter: JavaFutureToTwitterFutureConverter
+  ): Bijection[TwitterFuture[A], JavaFuture[A]] = {
+    new AbstractBijection[TwitterFuture[A], JavaFuture[A]] {
+      override def apply(f: TwitterFuture[A]): JavaFuture[A] =
+        f.toJavaFuture.asInstanceOf[JavaFuture[A]]
```
BenFradet, line 115: Since I can't use `CompletableFuture` I found this way around, I don't know if there is a better way, feedback welcome.

BenFradet, line 115: For that matter, I don't understand why [toJavaFuture](https://github.com/twitter/util/blob/develop/util-core/src/main/scala/com/twitter/util/Future.scala#L1232) returns a future of some subtype of A and not just a future of A.

BenFradet:
The failing test doesn't concern this pr:

> StringBijectionLaws:
> - round trips string -> Array[String] **\* FAILED ***

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -79,6 +81,22 @@ trait UtilBijections {
   }
 
   /**
+   * Injection from java futures to twitter futures
+   */
+  implicit def twitter2JavaFuture[A](
```
mosesn: it might make sense to expose an `injection` which doesn't need a converter too, in case folks are only going `JFuture => TwitterFuture`.

bijection-util/src/test/scala/com/twitter/bijection/twitter_util/UtilBijectionLaws.scala
```diff
@@ -16,72 +16,92 @@ limitations under the License.
...
+    TwitterTry(ScalaAwait.result(f, Duration.Inf)).toOption
 
-  implicit def futureArb[T: Arbitrary] = arbitraryViaFn[T, TwitterFuture[T]] { TwitterFuture.value(_) }
+  protected def toOption[T](f: JavaFuture[T]): Option[T] =
+    TwitterTry(f.get()).toOption
+
+  implicit def futureArb[T: Arbitrary] = arbitraryViaFn[T, TwitterFuture[T]] { TwitterFuture.value }
   implicit def scalaFutureArb[T: Arbitrary] = arbitraryViaFn[T, ScalaFuture[T]] { future(_) }
+  implicit def javaFutureArb[T: Arbitrary] =
+    arbitraryViaFn[T, JavaFuture[T]] { Futures.immediateFuture }
```
mosesn: I'm not sure this is worth entailing the dependency.  How do you feel about rolling your own with `FutureTask`?

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
BenFradet: You mean going `TwitterFuture => JavaFuture`?
In any case, I agree.

bijection-util/src/test/scala/com/twitter/bijection/twitter_util/UtilBijectionLaws.scala
BenFradet: I'll look into it.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
mosesn: Ah, yeah, my mistake.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
mosesn, line 115: I think it's because \forall B <: A, TwitterFuture[B] <: TwitterFuture[A], but the same isn't true for JFuture, since they encode covariance differently.

BenFradet, line 115: Ah ok, thanks for explaining

bijection-util/src/test/scala/com/twitter/bijection/twitter_util/UtilBijectionLaws.scala
```diff
@@ -16,72 +16,98 @@ limitations under the License.
...
   implicit def scalaFutureArb[T: Arbitrary] = arbitraryViaFn[T, ScalaFuture[T]] { future(_) }
+  implicit def javaFutureArb[T: Arbitrary] = arbitraryViaFn[T, JavaFuture[T]] { t =>
+    new JavaFuture[T] {
+      override def isCancelled: Boolean = false
+      override def get(): T = t
+      override def get(timeout: Long, unit: TimeUnit): T = t
+      override def cancel(mayInterruptIfRunning: Boolean): Boolean = false
+      override def isDone: Boolean = true
+    }
+  }
```
BenFradet, line 53: For some reason:

``` scala
new FutureTask[T](new Callable[T] {
  override def call(): T = t
}
```

and

``` scala
new FutureTask[T](new Runnable {
  override def run(): Unit = ()
}, t)
```

wouldn't work and take forever

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
BenFradet: However I'm not sure how I would go about inverting without a converter, ideas?

mosesn: injections only need to go one way.  you can just return None if it's not a ConstFuture.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -79,6 +81,44 @@ trait UtilBijections {
...
   /**
+   * Injection from twitter futures to java futures.
+   */
+  implicit def twitter2JavaFutureInjection[A]: Injection[TwitterFuture[A], JavaFuture[A]] = {
+    new AbstractInjection[TwitterFuture[A], JavaFuture[A]] {
+      override def apply(f: TwitterFuture[A]): JavaFuture[A] =
+        f.toJavaFuture.asInstanceOf[JavaFuture[A]]
+
+      override def invert(f: JavaFuture[A]): ScalaTry[TwitterFuture[A]] =
+        Inversion.attemptWhen(f)(_.isDone)(jf => TwitterFuture.value(jf.get()))
```
BenFradet: @mosesn I tried this approach, could you tell me what you think? thanks

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/JavaFutureConverter.scala
```diff
@@ -0,0 +1,81 @@
...
+        javaFuture.cancel(mayInterruptIfRunning)
+      }
+    }
+    p
+  }
+}
+
+/**
+ * Converter based on a [[Timer]] which will create a task which will check every
+ * <code>checkFrequency</code> if the java future is completed, one thread will be used for every
```
mosesn: the threading model is determined by the timer, not be the implementation.  most timers have a small, bounded thread pool, and they typically default to a size 1 thread pool.

it might also be worth noting that the minimum `checkFrequency` will be affected by the granularity of the timer.  If you're using a timer with tick granularity of 10ms, setting a `checkFrequency` of 0.5ms will not do you a ton of good.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/JavaFutureConverter.scala
```diff
@@ -0,0 +1,81 @@
...
+ *                              running
+ */
+class FuturePoolJavaFutureConverter(mayInterruptIfRunning: Boolean) extends JavaFutureConverter {
+  override def apply[T](javaFuture: JFuture[T]): Future[T] = {
+    val f = FuturePool.unboundedPool { javaFuture.get() }
+    val p = Promise.attached(f)
+    p.setInterruptHandler { case NonFatal(e) =>
+      if (p.detach()) {
+        f.raise(e)
+        javaFuture.cancel(mayInterruptIfRunning)
```
mosesn, line 49: we need to fail `p` too.  `p.setException(e)` should be sufficient here.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/JavaFutureConverter.scala
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Base class for converting java futures to twitter futures
+ */
+abstract class JavaFutureConverter {
+  def apply[T](javaFuture: JFuture[T]): Future[T]
+}
+
+/**
+ * Converter based on a [[FuturePool]] which will create one thread per future.
+ * To favor if there aren't too many futures to convert and one cares about latency.
```
mosesn, line 33: maybe worth adding a note that we use an unbounded threadpool to avoid potential deadlocks.

edit: thinking about this more, I think we're not at risk of deadlocks because of the way java futures work, so we might want to let people pass in their own `FuturePool` so they can provide a bound on the size of the thread pool if necessary.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
mosesn: LGTM

bijection-util/src/test/scala/com/twitter/bijection/twitter_util/UtilBijectionLaws.scala
mosesn, line 53: @BenFradet did you try calling `run` on the task after you created it?

BenFradet, line 53: yup, you were right :flushed:

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/JavaFutureConverter.scala
BenFradet: Out of curiosity, how do you find out the granularity of the timer you're using?

mosesn: read the docs :(.  you're right though, it might be useful to expose it programmatically.

briefly: j.u.Timer, 1 millisecond, j.u.c.ScheduledThreadPoolExecutor, 1 millisecond, netty HashedWheelTimer is configurable (1.ms minimum), finagle default timer uses netty HWT under the hood, and configures 10ms by default.

BenFradet: I thought you meant twitter util's Timers.

mosesn: @BenFradet I do, but there are many underlying implementations for the twitter util Timers.  In particular, we have support for j.u.Timer, j.u.c.ScheduledThreadPoolExecutor, netty HWT, and our own mocked out timer we can manipulate time for.

BenFradet: My bad then, I skimmed through [Timer.scala](https://github.com/twitter/util/blob/develop/util-core/src/main/scala/com/twitter/util/Timer.scala) too rapidly.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/JavaFutureConverter.scala
```diff
@@ -0,0 +1,89 @@
...
+class TimerJavaFutureConverter(
+  timer: Timer,
+  checkFrequency: Duration,
+  mayInterruptIfRunning: Boolean
+) extends JavaFutureConverter {
+  override def apply[T](javaFuture: JFuture[T]): Future[T] = {
+    val p = Promise[T]
+    val task = timer.schedule(checkFrequency) {
+      if (javaFuture.isDone) {
+        p.updateIfEmpty(Try(javaFuture.get()))
```
johnynek, line 80: do we not need to unschedule this task once completed?

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
johnynek: don't you want `TwitterFuture(jf.get)` since `.get` can throw?

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -79,6 +81,44 @@ trait UtilBijections {
   }
 
   /**
+   * Injection from twitter futures to java futures.
```
johnynek, line 84: can we add some comment that the invert here fails on any java future that is not done? This is a bit surprising to have both an `Injection` and `Bijection` between two types. I don't think this weak one should be implicit, if you ask me. It can be here without the `implicit`, but it is a bit dangerous to use.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -79,6 +81,44 @@ trait UtilBijections {
...
+   *   <code>checkFrequency</code>.
+   */
+  implicit def twitter2JavaFutureBijection[A](
+    implicit converter: JavaFutureConverter
+  ): Bijection[TwitterFuture[A], JavaFuture[A]] = {
+    new AbstractBijection[TwitterFuture[A], JavaFuture[A]] {
+      override def apply(f: TwitterFuture[A]): JavaFuture[A] =
+        f.toJavaFuture.asInstanceOf[JavaFuture[A]]
+
+      override def invert(f: JavaFuture[A]): TwitterFuture[A] = {
```
johnynek: no need for `{ }` here

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/JavaFutureConverter.scala
BenFradet, line 80: I think we do, I haven't found out how yet.
If it'd had been a j.u.Timer we would have been able to do:

``` scala
if (javaFuture.isDone) {
  p.updateIfEmpty(Try(javaFuture.get()))
  timer.cancel()
  timer.purge()
}
```

But c.t.u.Timer doesn't expose the same interface.

mosesn, line 80: we need to call `task.cancel()`.  you might need to mutate a reference.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
BenFradet, line 84: Agreed to remove the implicit and for the doc.
I still think it's useful based on @mosesn remark that if someone needs only to go `TwitterFuture` => `JavaFuture` it makes sense to expose something which doesn't take any useless implicits.

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/JavaFutureConverter.scala
BenFradet, line 80: @mosesn I'm not sure to get what you mean.

mosesn, line 80: Here's an example:

``` scala
var task = null
task = timer.schedule(...) {
  ...
  task.cancel()
}
```

johnynek, line 80: you could do
`lazy val task = timer.schedule(...)  { ... task.cancel() }`

lazy val can be recursive. I prefer that to the var here.

BenFradet, line 80: I was afraid the first solution wouldn't work / be too awful.

I'll try the second one, didnt know this could be possible, thanks.

johnynek, line 80: you'll have to force the `task` after though. Something like `require(task != null)` to actually schedule it, which is also a bit janky.

BenFradet, line 80: indeed :'(

johnynek:
Thank you!

BenFradet:
Thanks a lot @johnynek @mosesn for your great help!

PR #249, opened 2016-06-12
Add support for Tuple1 bijections

Olshansk:
- Add support for Bufferable injection from Tuple1[T] to Array[Byte]
- Add support for injection from a case class with one element to a
  Tuple1[T]

Testing:
$ ./sbt bijection-macros/console

scala> import com.twitter.bijection.{Bijection, Bufferable, Injection, macros}
scala> import macros._
scala> import MacroImplicits._
scala> case class Foo(v: Long)
scala> implicit val fooToByteArray = Bufferable.injectionOf[Tuple1[Long]]
scala> val fooInjection = Injection.connect[Foo, Tuple1[Long], Array[Byte]]


johnynek:
Thanks for the improvement!

PR #250, opened 2016-08-03
Add mima checks to travis

johnynek:
We already made a few minor breaks, so the next version should be `0.10.0` (although probably no one would get hit by the differences so far).


ianoc:
LGTM

rubanm:
👍

PR #251, opened 2016-08-03
try to setup codecov.io

johnynek:
I _think_ maybe twitter already has some codecov.io, so this may just work.


.travis.yml
```diff
@@ -6,4 +6,6 @@ matrix:
       script: ./sbt ++$TRAVIS_SCALA_VERSION clean test mimaReportBinaryIssues
 
     - scala: 2.11.8
-      script: ./sbt ++$TRAVIS_SCALA_VERSION clean test mimaReportBinaryIssues
+      script: ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
```
rubanm, line 9: Are we doing this for 2.11 only because the results should be identical for 2.10?

rubanm:
👍

.travis.yml
johnynek, line 9: yeah. That was my thinking.

PR #252, opened 2016-08-25
Add scalafmt

johnynek:
This turns on scalafmt as a test so you can't merge the code unless your code is clean with respect to scalafmt.

`sbt scalafmt`

runs the formatter.

`sbt scalafmtTest` runs just the test (which is run on travis before any tests).


isnotinvain:
👍

PR #256, opened 2016-09-20
Avro - toJson should work exactly as toBinary

mariussoutier:
Adapt `toJson` to work similarly as `toBinary` in order to support generated Scala case classes as specific records.

### DIFF
--- a/bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
+++ b/bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
@@ -103,6 +103,18 @@ object SpecificAvroCodecs {
 
   /**
     * Returns Injection capable of serializing and deserializing a compiled avro record using org.apache.avro.io.BinaryEncoder.
+    * @tparam T compiled Avro record
+    * @param schema The record's schema
+    * @return Injection
+    */
+  def toBinary[T <: SpecificRecordBase](schema: Schema): Injection[T, Array[Byte]] = {
+    val writer = new SpecificDatumWriter[T](schema)
+    val reader = new SpecificDatumReader[T](schema)
+    new BinaryAvroCodec[T](writer, reader)
+  }
+
+  /**
+    * Returns Injection capable of serializing and deserializing a compiled avro record using org.apache.avro.io.BinaryEncoder.
     * Fetches the schema from the specified class in order to be compatible with generated Scala classes.
     * @tparam T compiled Avro record
     * @return Injection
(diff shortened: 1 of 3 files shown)


bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
```diff
@@ -121,10 +121,11 @@ object SpecificAvroCodecs {
     * @tparam T compiled Avro record
     * @return Injection
     */
-  def toJson[T <: SpecificRecordBase: ClassTag](schema: Schema): Injection[T, String] = {
-    val klass = classTag[T].runtimeClass.asInstanceOf[Class[T]]
-    val writer = new SpecificDatumWriter[T](klass)
-    val reader = new SpecificDatumReader[T](klass)
+  def toJson[T <: SpecificRecordBase: ClassTag]: Injection[T, String] = {
```
johnynek, line 149: can we add a new method name and not break binary compatibility?

johnynek (changes requested):

bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
mariussoutier: Sure, it's what I suggested when we changed `toBinary`, but you guys decided it was unnecessary. I'll update the PR.

johnynek:
We since added a binary compatibility check to CI and changed the process
for publishing when that check fails. Sorry for the hassle.

mariussoutier:
Makes sense. Is the updated PR not binary compatible? 

Edit: Ah I think there's a formatting error, but I'm not sure what it is.

johnynek:
Failed the scala format test. Can you run sbt scalafmt and commit the changes?

mariussoutier:
Done!

bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
```diff
@@ -103,30 +103,53 @@ object SpecificAvroCodecs {
...
   }
 
   /**
     * Returns Injection capable of serializing and deserializing a generic avro record using org.apache.avro.io.JsonEncoder to a
     * UTF-8 String
     * @tparam T compiled Avro record
+    * @param schema The record's schema
     * @return Injection
     */
   def toJson[T <: SpecificRecordBase: ClassTag](schema: Schema): Injection[T, String] = {
```
johnynek, line 136: I'd live to remove the `ClassTag` here if possible, but that would be a binary incompatible break, so we can leave it as a TODO for now.

bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
```diff
@@ -103,30 +103,53 @@ object SpecificAvroCodecs {
 
   /**
     * Returns Injection capable of serializing and deserializing a compiled avro record using org.apache.avro.io.BinaryEncoder.
+    * @tparam T compiled Avro record
+    * @param schema The record's schema
+    * @return Injection
+    */
+  def toBinary[T <: SpecificRecordBase: ClassTag](schema: Schema): Injection[T, Array[Byte]] = {
```
johnynek: does this actually use the `ClassTag`? I don't see where. If not, since we already dropped it in a previous version, why not correct that issue?

johnynek (changes requested):
If we can, let's remove the `ClassTag` on the `toBinary`. If I'm missing something, let's add a comment and leave it.

Thanks for updating the PR!

bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
mariussoutier: Good point, done!

bijection-avro/src/main/scala/com/twitter/bijection/avro/AvroCodecs.scala
mariussoutier: Ok I added a todo.

johnynek (approved):
Thanks!

mariussoutier:
All checks successful, merge away ;)

johnynek:
Thank you!

PR #257, opened 2016-11-26
Clean Up Build To Clear The Path to 2.12

clhodapp:
I'm anxious to see finagle released for Scala 2.12, so I decided to try to help unblock bijection's 2.12 release. This is my attempt to clean up the build, updating it to sbt 0.13.13 and cleaning up deprecated cruft. It was my explicit intention to NOT do anything that would have a binary compatibility impact here (i.e. this does NOT actually enable 2.12 builds for bijection). By keeping things separate, it should be easier to verify if I've broken things with this build refactoring. In theory, this shouldn't change anything about the generated artifacts (modulo maybe some harmless changes to the pom, since I've moved some things off of the gross `pomExtra` and onto conventional sbt settings). I _did_ update build and test dependencies, since those shouldn't affect the released artifacts.

I will note that I can't really test that I didn't break the release process, so sorry in advance if this gets merged and then someone has to fix it.


build.sbt
```diff
@@ -1,156 +1,153 @@
...
-      "-language:implicitConversions",
-      "-language:higherKinds",
-      "-language:existentials",
-      "-Xmax-classfile-name",
-      "200"), // People using encrypted file-systems can have problems if the names get too long
+val buildLevelSettings = Seq(
+  organization := "com.twitter",
+  scalaVersion := "2.11.8",
+  crossScalaVersions := Seq("2.10.6", "2.11.8"),
+  scalaVersion := "2.11.8",
```
xuwei-k, line 11: duplicate?

xuwei-k (changes requested):

build.sbt
clhodapp: Good catch

johnynek:
Thanks a ton! As you can see, none of us are sbt experts (we are barely sbt users ;).

piyushnarang:
Thanks for taking a stab at this @clhodapp. I was working on enabling 2.12 for bijection but got stuck on some the failing HBase tests (which you've fixed). I'll try and get a review out to turn on 2.12 publishes and get some artifacts out today.

PR #258, opened 2016-11-28
Add support for Scala 2.12.0

piyushnarang:
Building on https://github.com/twitter/bijection/pull/257. Adding 2.12.0 as a cross version to include. Had to disable building 2.12 for a few projects as they require some finagle 2.12 artifacts (which in turn requires bijection-core 2.12). Once the finagle artifacts are in, we can reenable those projects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,6 @@ matrix:
 
     - scala: 2.12.0
       jdk: oraclejdk8
-      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
+      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION test"
       after_success:
         - bash <(curl -s https://codecov.io/bash)
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ matrix:
       script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION coverage" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION coverageReport" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
 
     - scala: 2.12.0
-          script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION coverage" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION coverageReport" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
+      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION coverage" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION coverageReport" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
       after_success:
         - bash <(curl -s https://codecov.io/bash)
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,12 @@ sudo: false
 matrix:
   include:
     - scala: 2.10.6
-      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean"  "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
+      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION scalafmtTest" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
 
     - scala: 2.11.8
-      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION coverageReport" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
+      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION scalafmtTest" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION coverageReport" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
 
     - scala: 2.12.0
-      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION coverageReport" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
+      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION scalafmtTest" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION coverageReport" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
       after_success:
         - bash <(curl -s https://codecov.io/bash)
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,12 @@ sudo: false
 matrix:
   include:
     - scala: 2.10.6
-      script: ./sbt ++$TRAVIS_SCALA_VERSION clean scalafmtTest test mimaReportBinaryIssues
+      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean"  "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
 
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -3,9 +3,12 @@ sudo: false
 matrix:
   include:
     - scala: 2.10.6
-      script: ./sbt ++$TRAVIS_SCALA_VERSION clean scalafmtTest test mimaReportBinaryIssues
+      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean"  "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
```
johnynek: why the 3 `+++`? I thought it was just `++`?

piyushnarang: Had to hook up the sbt-doge plugin to ensure that the overridden crossScalaVersions is respected. Without this I still get the error in the 3 projects we want to skip. I think we've done something similar in: https://github.com/twitter/chill/pull/253/

clhodapp:
FWIW, `scalafmtTest` _should_ work. The output is a bit janky (there seems to be some mixing between the output for different projects b/c of the way I hacked calling it into the build), but it should report accurate results.

piyushnarang:
@clhodapp ah you're right. Just tried running it and it seems to work. I can update to turn it back on.

clhodapp (approved):
Nothing here stands out as crazy

johnynek:
If we turn on coverage by default doesn't that dramatically slow down the local tests?

piyushnarang:
@johnynek looking at the build time in travis - https://travis-ci.org/twitter/bijection/builds/179597059
Was around 5-6 mins. There seems to be a fair bit of variance in general in the builds there. Seems like it was turned on earlier for 2.11.8 right? Do we want to have it turned on only for that?

johnynek:
you should not need to do coverage for more than one version of scala, since it is the same code, right?

I think just picking one version "the main one", whatever we consider that to be, should work.

piyushnarang:
@johnynek yes, running coverage on one version should suffice. I wasn't able to figure out how to enable coverage via the command line using sbt. Was running into an error:
```
[info] Set current project to bijection (in build file:/home/travis/build/twitter/bijection/)
[error] Expected ':' (if selecting a configuration)
[error] Not a valid key: coverage (similar: coverageReport, coverageEnabled, coverageMinimum)
[error] bijection-core/coverage
[error]                        ^
The command "./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION coverage" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION coverageReport" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"" exited with 1.
```

Can check if there's an issue running the sbt-coverage plugin with the sbt-doge `+++` operator. Another option would be to enableCoverage in the build.sbt only for 2.11.8.

piyushnarang:
@johnynek ok not 100% sure what the best way to proceed on this is. Here's what I see. When I set up coverageEnabled conditionally:
````
val buildLevelSettings = Seq(
...
coverageEnabled := (if (scalaVersion.value startsWith "2.11") true else false),
...
)
````

I still end up with coverage running in 2.12.0:
```
[tw-mbp-pnarang bijection (scala-2.12)]$ ./sbt "+++2.12.0 test"
[info] Loading global plugins from /Users/pnarang/.sbt/0.13/plugins
[info] Loading project definition from /Users/pnarang/workspace/bijection/project
[info] Resolving key references (10437 settings) ...
[info] Set current project to bijection (in build file:/Users/pnarang/workspace/bijection/)
[info] Setting version to 2.12.0
[info] Reapplying settings...
[info] Set current project to bijection (in build file:/Users/pnarang/workspace/bijection/)
[info] Compiling 27 Scala sources and 2 Java sources to /Users/pnarang/workspace/bijection/bijection-core/target/scala-2.11/classes...
[info] [info] Cleaning datadir [/Users/pnarang/workspace/bijection/bijection-core/target/scala-2.11/scoverage-data]
[info] [info] Beginning coverage instrumentation
...
```
 Seems like our default setting of scalaVersion = 2.11.8 is ensuring that this can't be updated even if we try to pass it with `++` or `+++` (it always is 2.11.8). 

This shows up even in the sbt console:
```
$ ./sbt
> ++ 2.10.6
[info] Setting version to 2.10.6
[info] Reapplying settings...
[info] Set current project to bijection (in build file:/Users/pnarang/workspace/bijection/)
> show scalaVersion
[info] bijection-guava/*:scalaVersion
[info] 	2.11.8
...
```

Now if I comment out the line in build.sbt setting `scalaVersion := "2.11.8"`, this seems to work ok - both the sbt console and the coverage builds behave as expected. We show the version passed in `++` or `+++` and coverageEnabled = true only for 2.11. 

Seems like this problem is present even currently. We are checking scalaVersion to set scalacOptions - https://github.com/twitter/bijection/blob/develop/build.sbt#L26. When I tested this out on the sbt command line, it seems like we're always setting the options for 2.11. 

My proposal is to remove the line setting `scalaVersion`. There might be a better way to do this that I may be missing so if folks have other ideas / suggestions, I could try them out.

**Comment by clhodapp** [comment]
What if you move the setting for `coverageEnabled` down into the `module` method?

File: build.sbt [hunk 2]
```diff
@@ -7,10 +7,11 @@ import bijection._
 
 val buildLevelSettings = Seq(
   organization := "com.twitter",
-  crossScalaVersions := Seq("2.10.6", "2.11.8"),
+  crossScalaVersions := Seq("2.10.6", "2.11.8", "2.12.0"),
   scalaVersion := "2.11.8",
   javacOptions ++= Seq("-source", "1.6", "-target", "1.6"),
   javacOptions in doc := Seq("-source", "1.6"),
+  coverageEnabled := true,
```
johnynek: do we need to enable this in the build? Can't we do `+++$TRAVIS_SCALA_VERSION coverage test` or something?

What is it about this change that is requiring us to always do coverage?

clhodapp: That's a good point. As a dev just trying to see if you broke the tests, you probably wouldn't want coverage to run anyway.

piyushnarang: Yeah so I originally tried to do that. The `+++$TRAVIS_SCALA_VERSION coverage` command doesn't work with the sbt-doge plugin for some reason (see above): https://github.com/twitter/bijection/pull/258#issuecomment-263434782

I needed to pull in the sbt-doge plugin cause I wanted to ensure that bijection-scrooge, bijectionUtil and bijectionFinagleMySql can override the cross-scala versions to skip 2.12 as we don't have the finagle artifacts. Based on searching that seemed to be the best option to skip modules selectively for a given scala version. If you folks know of a better alternative, I could try that and drop the doge plugin.

piyushnarang:
@clhodapp / @johnynek - made some updates, now code coverage is enabled only on 2.11. Some of the pain is related to sbt-doge which doesn't seem to have been working with the `inThisBuild` method for some reason. It was always using scalaVersion = 2.11.8. The sbt-doge changes are temporary till finagle is able to publish their 2.12 jars. At that point we can yank it out and enable the builds on the 3 projects that depend on finagle.

clhodapp:
Not setting a build-level scala version breaks tools like ensime.

Could you try moving just `scalaVersion` out of the list as `scalaVersion in ThisBuild := "2.11.8"` and renaming the list back to `commonSettings`? I wasn't thinking about compatibility with `sbt-doge` when I did my build refactoring, but I believe that doing this should work.

piyushnarang:
@clhodapp so I tried setting the scalaVersion like you specified and renaming. Seems like that doesn't end up compiling to 2.12. So the version ends up staying at 2.11.8 even though we specify +++2.12.0. 
I don't think we were setting the build level scala version earlier right? I guess I'm also wondering if it is ok to break it for just the finagle release to proceed? Once the finagle jars are published, we don't need the sbt-doge stuff.

clhodapp:
Bleh. OK. I give up then haha. With your PR as-is, bijection is no worse off in any way than it was a few days ago, so it seems reasonable. Once you remove `doge`, you can set a build-level scala version if desired.

johnynek:
Yeah, I guess that's fine. It's a bummer to make the tests always slow, but
I guess doge is incompatible with coverage somehow.

mosesn (approved):

.travis.yml
```diff
@@ -3,9 +3,13 @@ sudo: false
 matrix:
   include:
     - scala: 2.10.6
-      script: ./sbt ++$TRAVIS_SCALA_VERSION clean scalafmtTest test mimaReportBinaryIssues
+      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION scalafmtTest" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
```
sritchie, line 6: one last thing - do we want this binary check for 2.10 as well? In algebird we only check the 2.11 build.

sritchie:
Other than my one comment, 👍 !

johnynek:
Actually good point @sritchie but I do think we want to check binary compat in all versions. For instance, the rules in scala 2.12 are different than 2.11 (more changes should be binary compatible due to java 8 features being used). Scala 2.10 and 2.11 might be the same or close, but just to be sure it seems like a good idea to run on all supported versions.

piyushnarang:
Thanks for the review and feedback @johnynek, @clhodapp, @mosesn, @sritchie! Will follow up once the finagle artifacts are out with cleaning this back up :-)

stuhood:
Thanks everyone! Will watch for a release.

sritchie:
@stuhood bijection is out! 0.9.4 should get you going. We're working on algebird now.

PR #261, opened 2016-12-03
add custom jvm opts

sritchie:
algebird build was way faster with these.


johnynek:
If the build is significantly faster, :+1: , let's just check and add a comment here recording the diff for history.

sritchie:
Not a big change.

PR #262, opened 2016-12-10
Fix up build, get bijection-util onto 2.12

sritchie:
This PR:

- Removes the always-enabled-coverage from the build (we don't need `sbt-doge` for 2.11 or 2.10, so we can use the old style in `.travis.yml`
- formats plugins
- upgrades `util-core` for 2.12 only
- adds caching to the build
- adds better badges to README.md
- adds a notification webhook for gitter to the .travis.yml
 - pulls some versions out into `val`s
- removes old doc generation (I'll make a microsite soon)


sritchie:
Cc @johnynek

build.sbt
```diff
@@ -249,8 +261,8 @@ lazy val bijectionScrooge = {
     libraryDependencies ++= Seq(
       "org.apache.thrift" % "libthrift" % "0.6.1" exclude ("junit", "junit"),
```
johnynek, line 262: I wonder if we could move this to util now. @mosesn? Seems like a module there, and/or finagle is the right place.

johnynek: I guess this is for Scrooge, but you get the idea.

johnynek (approved):

build.sbt
mosesn: DM me and we can chat about it.

PR #263, opened 2017-01-07
Add support to publish all bijection modules with Scala 2.12

piyushnarang:
As the finagle 2.12 jars are out, picking them up and publishing all bijection modules in 2.12. Finagle has stopped supporting 2.10 (http://finagle.github.io/blog/2016/04/20/scala-210-and-java7/) and they've since moved a bunch of packages around. I've turned off building / publishing bijectionFinagleMySql for 2.10 to get around this. We could also go the finagle route and stop building / publishing for 2.10. 

Also made a couple of changes to some test code to get around some code in ScalaRuntime being dropped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ matrix:
       script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
 
     - scala: 2.11.8
+      jdk: oraclejdk8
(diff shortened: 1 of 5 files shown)


johnynek:
`java.lang.UnsupportedClassVersionError: com/twitter/finagle/mysql/TimestampValue : Unsupported major.minor version 52.0`

you need to tell travis to use oraclejdk8 for 2.11 and 2.12 now, it looks like.

piyushnarang:
@johnynek yeah I just updated that. I've already switched on my machine so didn't run into it :-)

johnynek:
+1 from me.

piyushnarang:
Thanks @johnynek. I'll cut a new release and publish jars soon.

johnynek:
👍

PR #268, opened 2017-08-30
make twitter util, finagle, scrooge provide dependencies

piyushnarang:
Bumping up the version of utils from 6.39.0 to 7.0.0 for Scala 2.11 & 2.12 at the request of @cacoco as they are seeing some issues with projects pulling in bijection with the older utils version and other libs with utils 7.0.0 (I'll let @cacoco expand on that if needed). 
@johnynek - wdyt?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,30 +6,30 @@ import ReleaseTransformations._ // for sbt-release.
 import bijection._
 
 val finagleVersion210 = "6.35.0"
-val finagleVersion = "7.0.0"
+val finagleVersion = "6.41.0"
 
 val scalatestVersion = "3.0.1"
 val scalacheckVersion = "1.13.4"
 
 val utilVersion210 = "6.34.0"
-val utilVersion = "7.0.0"
+val utilVersion = "6.39.0"
 
 val scroogeSerializerVersion210 = "3.17.0"
-val scroogeSerializerVersion = "4.19.0"
+val scroogeSerializerVersion = "4.13.0"
 
 def util(mod: String, scalaVersion: String) = {
   val version = versionFallback(scalaVersion, utilVersion210, utilVersion)
-  "com.twitter" %% (s"util-$mod") % version
+  "com.twitter" %% (s"util-$mod") % version % "provided"
 }
 
 def finagle(mod: String, scalaVersion: String) = {
   val version = versionFallback(scalaVersion, finagleVersion210, finagleVersion)
-  "com.twitter" %% (s"finagle-$mod") % version
+  "com.twitter" %% (s"finagle-$mod") % version % "provided"
 }
 
 def scroogeSerializer(scalaVersion: String) = {
   val version = versionFallback(scalaVersion, scroogeSerializerVersion210, scroogeSerializerVersion)
-  "com.twitter" %% "scrooge-serializer" % version
+  "com.twitter" %% "scrooge-serializer" % version % "provided"
 }
 
 def versionFallback(scalaVersion: String, packageVersion210: String, version: String) = {
@@ -280,7 +280,7 @@ lazy val bijectionScrooge = {
       "org.apache.thrift" % "libthrift" % "0.6.1" exclude ("junit", "junit"),
       scroogeSerializer(scalaVersion.value),
       util("core", scalaVersion.value),
-      finagle("core", scalaVersion.value) % "test"
+      finagle("core", scalaVersion.value)
     )
   ).dependsOn(
     bijectionCore % "test->test;compile->compile",
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ import ReleaseTransformations._ // for sbt-release.
 import bijection._
 
 val finagleVersion210 = "6.35.0"
-val finagleVersion = "6.41.0"
+val finagleVersion = "7.0.0"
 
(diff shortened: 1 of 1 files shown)


cacoco (dismissed):
I believe you should also rev the finagle and scrooge versions as all of these should be consistent with each other. Finagle is also now 7.0.0 and scrooge is now 4.19.0.

piyushnarang:
@cacoco - updated finagle to 7.0.0 and scrooge to 4.19.0

johnynek:
What's the binary issue @cacoco ?
 
If there is not a binary issue, this upgrade pushes people up. We use this code at stripe but we are on a fairly old util/finagle (due to GC issues we saw one of the finagle bumps which we have not spent the time to debug yet).

My general feeling is that dependencies should be their oldest ones still binary compatible with the current one so we don't impose too much pain downstream.

cacoco:
@johnynek Finatra has a dependency on the latest published version of bijection-util, 0.9.5. After the latest release of CSL OSS libraries we got a couple of users through Gitter who had hit some classpath issues after upgrading Finatra since Finatra also has it's own dependency on util-core which was updated to 7.0.0 but bjiection-util (through bijection-core) still has a dependency on util-core 6.39.0. I simply asked the bijections team if it was possible to get a new release against the lastest versions of finagle, scrooge and util to minimize issues for users.

If this is not something you want to do, that's totally fine. 

We've since added an exclusion on the bijection-util dependency to exclude util-core but it doesn't look like the version of sbt we're using really cares (still seems to get pulled in an then evicted -- at least it gets evicted for our example projects). I'm guessing it's because it's a second-order transitive dependency. 

Thanks!

johnynek:
I think I understand Chris: the issue is the eviction warning, not an actual run-time or compile-time failure, yes?

If that's the case, I think we could mark `twitter-util` as `provided`, which it probably should be. In that way, the user of bijection-util is expected to bring their own version of `twitter-util`, which makes sense. This is an add-on, it shouldn't be the thing setting your dependency on a util version.

What do you think about that?

mosesn:
TIL about provided scope!  Very cool.

cacoco:
@johnynek if you're cool with your users needing to bring a `util-core` to the table, then I think provided scope would work just fine. 

However, the `util-core` dependency shows up when adding a dependency on `bijection-util` through its dependency on `bijection-core` which depends on `util-core`. Are you saying you'd mark the dependency on `util-core` in `bijection-core` as provided? 

Since `bijection-util` has no direct dependency on `util-core` there would be no way to mark it provided there, so it'd have to be on `bijection-core`, right? I ask because I want to make sure when say "this is an add-on" that you meant specifically `bijection-util` or did you mean all of bijection?

Thanks!

mosesn:
@cacoco where do you see that dependency?  It looks like it's through bijection-util to me.

[bijection-core](https://github.com/twitter/bijection/blob/develop/build.sbt#L210-L233)
```
/** No dependencies in bijection other than java + scala */
lazy val bijectionCore = {
  module("core").settings(
    osgiExportAll("com.twitter.bijection"),
    libraryDependencies ++= Seq(
      "com.novocode" % "junit-interface" % "0.11" % "test"
    ),
    sourceGenerators in Compile += Def.task {
      val main = (sourceManaged in Compile).value
      val out = streams.value
      val pkg = main / "scala" / "com" / "twitter" / "bijection"
      def genSrc(name: String, gen: => String) = {
        val srcFile = pkg / name
        IO.write(srcFile, gen)
        out.log.debug(s"generated $srcFile")
        srcFile
      }
      Seq(
        genSrc("GeneratedTupleBijections.scala", Generator.generate),
        genSrc("GeneratedTupleBuffer.scala", BufferableGenerator.generate)
      )
    }.taskValue
  )
}
```

[bijection-util](https://github.com/twitter/bijection/blob/develop/build.sbt#L301-L308)
```
lazy val bijectionUtil = {
  module("util").settings(
    osgiExportAll("com.twitter.bijection.twitter_util"),
    libraryDependencies += util("core", scalaVersion.value)
  ).dependsOn(
    bijectionCore % "test->test;compile->compile"
  )
}
```

cacoco:
@mosesn running dependency-tree on finatra shows bijection-util --> bijection-core --> util-core.

johnynek:
@cacoco there is a bug somewhere, because bijection-core should have no dependencies (as indicated above).

as you see here:
https://search.maven.org/remotecontent?filepath=com/twitter/bijection-core_2.12/0.9.5/bijection-core_2.12-0.9.5.pom

the only dependencies are the scala-library and test dependencies.

cacoco:
@johnynek ok -- more than happy to know that its wrong. The provides should be good, then.

johnynek:
@piyushnarang can you change to make all the twitter libs `provided`? finagle, util, scrooge?

piyushnarang:
@johnynek yeah I can make the update today. I noticed that @jensraaby has a PR out too - https://github.com/twitter/bijection/pull/266.

johnynek:
👍 
Looks good to me. Should fix the issue.

johnynek (approved):

piyushnarang:
@cacoco - wdyt?

cacoco:
@piyushnarang looks good!

piyushnarang:
Thanks for taking a look @johnynek / @cacoco

PR #287, opened 2019-08-22
cross-build for 2.13

martijnhoekstra:


martijnhoekstra:
At least bijection-core

nevillelyh:
@martijnhoekstra are you still working on this? I can pick it up if you don't mind.

martijnhoekstra:
Oh, yes, I can pick this up again; I thought this  one was ready to go, but maybe it isn't.

build.sbt
```diff
@@ -39,7 +39,7 @@ def versionFallback(scalaVersion: String, packageVersion210: String, version: St
 
 val buildLevelSettings = Seq(
   organization := "com.twitter",
-  crossScalaVersions := Seq("2.10.6", "2.11.8", "2.12.1"),
+  crossScalaVersions := Seq("2.10.6", "2.11.8", "2.12.1", "2.13.0"),
```
nevillelyh: Worth updating all of them? I think the latest are 2.10.6, 2.11.12, 2.12.10, & 2.13.1

build.sbt
```diff
@@ -8,14 +8,14 @@ import bijection._
...
 val finagleVersion = "6.41.0"
 
-val scalatestVersion = "3.0.1"
-val scalacheckVersion = "1.13.4"
+val scalatestVersion = "3.0.8"
+val scalacheckVersion = "1.14.0"
 
 val utilVersion210 = "6.34.0"
-val utilVersion = "6.39.0"
+val utilVersion = "19.9.0-SNAPSHOT"
```
nevillelyh: https://mvnrepository.com/artifact/com.twitter/util-core
Looks like 19.9.0 is out.

build.sbt
```diff
@@ -8,14 +8,14 @@ import bijection._
...
 
 val scroogeSerializerVersion210 = "3.17.0"
-val scroogeSerializerVersion = "4.13.0"
+val scroogeSerializerVersion = "19.9.0-SNAPSHOT"
```
nevillelyh: 19.9.0 is out for scrooge too.

build.sbt
```diff
@@ -8,14 +8,14 @@ import bijection._
 val finagleVersion210 = "6.35.0"
 val finagleVersion = "6.41.0"
```
nevillelyh: Latest finagle is also 19.9.0 now and built for scala 2.13. Should we use that?
https://mvnrepository.com/artifact/com.twitter/finagle-core

nevillelyh:
Also we should probably get the CI build green.
@johnynek worth adding to scala-steward in the future?

johnynek:
Yeah, I think scala-steward is a win for all repos.

build.sbt
martijnhoekstra: 19.10.0 should be on its way. I can put everything on that if desirable and we can hold off for a few days.

mosesn: indeed, 19.10.0 has been published, looks like mvnrepository might have a stale cache https://search.maven.org/artifact/com.twitter/finagle-core_2.13/19.10.0/jar

build.sbt
mosesn: 2.10.7, I think

martijnhoekstra:
https://github.com/scala-steward-org/repos/pull/28

martijnhoekstra:
Oh, it turns out I had lots of half-baked stuff laying around for bijection. There was lots of stuff that needed fixing, and I'm not sure what order I had in mind for that. Maybe first we can see what scala-steward brings to the table.

nevillelyh:
@martijnhoekstra what about a quick fixup on this PR so we can get it in and let scala-steward pick up the rest?

martijnhoekstra:
@nevillelyh if it indeed turns out to be a quick fixup I'm all for it. When I get home tonight I can work on that, and see if it is and the problem is just scalafmt, or if there is more going on.

martijnhoekstra:
quick update: this is far farther from home than I had hoped.

johnynek:
I'd like to use the strategy of making a `scala-2.12-` directory and `scala-2.13+` directory or something for any changes. Since we haven't published for 2.13, anything goes, but for 2.11/2.12 I'd really like to not break binary compatibility.

We did something like this here:
https://github.com/typelevel/paiges/pull/189

martijnhoekstra:
that sounds good @johnynek -- even without the bincompat argument it'll be necessary.

martijnhoekstra:
I burned this branch to the ground, and started it with a new blank slate. Marked the PR WIP.

bijection-core/src/test/scala/com/twitter/bijection/CollectionLaws.scala
```diff
@@ -22,26 +22,50 @@ import org.scalatest.MustMatchers
...
+    type CA = List[A]
+    type CB = Vector[B]
+
+    val ali = implicitly[Arbitrary[CA]]
+    val avsri = vectorArb(implicitly[Arbitrary[List[B]]])
+    //if we start here with Bijection.fromInjection(Injection.int2String)
+    //things do work out. Without, we get diverging expansion starting at 
+    //Injection.fromImplicitBijection
+    implicit val i1 = Bijection.fromInjection(Injection.int2String)
+    def ibij: ImplicitBijection[CA, CB] = implicitly[ImplicitBijection[List[Int], Vector[String @@ Rep[Int]]]]
```
martijnhoekstra: @johnynek I pushed my WIP, as I'm hitting a hurdle here: scala 2.13 runs out of heap trying to get some sort of LUB here (I haven't figured out what it's trying to LUB). Ideas/sage advice/sharp insight more than welcome.

johnynek: The whole `A @@ Rep[B]` trick is very rarely used. In general, most cases would use `Injection[A, B]` It may be that something with the tricks to do `@@` 

 https://github.com/twitter/bijection/blob/develop/bijection-core/src/main/scala/com/twitter/bijection/Rep.scala

https://github.com/twitter/bijection/blob/develop/bijection-core/src/main/scala/com/twitter/package.scala#L55

bijection-core/src/test/scala/com/twitter/bijection/CollectionLaws.scala
```diff
@@ -22,26 +22,50 @@ import org.scalatest.MustMatchers
...
+    type B = String @@ Rep[Int]
+    type CA = List[A]
+    type CB = Vector[B]
+
+    val ali = implicitly[Arbitrary[CA]]
+    val avsri = vectorArb(implicitly[Arbitrary[List[B]]])
+    //if we start here with Bijection.fromInjection(Injection.int2String)
+    //things do work out. Without, we get diverging expansion starting at 
+    //Injection.fromImplicitBijection
+    implicit val i1 = Bijection.fromInjection(Injection.int2String)
```
johnynek: it could help to put a type here, maybe `Bijection[Int, String @@ Rep[Int]]`?

martijnhoekstra: This keeps getting weirder. With 

```
implicit val i1 = Bijection.fromInjection(Injection.int2String)
def ibij: ImplicitBijection[CA, CB] = implicitly[ImplicitBijection[CA, CB]] 
```

it succeeds fine.

With `def ibij: ImplicitBijection[CA, CB] = implicitly[ImplicitBijection[List[Int], Vector[String @@ Rep[Int]]` it crashes the compiler with OOM. That's obviously a compiler bug, but it's not immediately clear to me if,  and if so how we can work around.

martijnhoekstra:
https://github.com/scala/bug/issues/11770 messes up bijection-clojure. Worked around by calling generated static java methods.

bijection-core/src/test/java/com/twitter/bijection/TestBijectionInJava.java
```diff
@@ -49,16 +49,18 @@ public void testBasicInjection() {
     //TODO include a cleaner way to get to the scala Bijections than Bijection$.MODULE$.
     @Test
     public void testBase64Bijection() {
-        // Note, value classes return the underlying types in Java. But Java users usually
-        // don't care much about type safety, so punting on this for now
-        Bijection<byte[], String> bytes2Base64 = Bijection$.MODULE$.bytes2Base64();
+        //https://github.com/scala/scala/pull/8127
+        //generic AnyVal signatures pre-2.12.9 incorrectly had the underlying
+        //type, while in reality it's the boxed AnyVal
+        //Bijection<byte[], Base64String> bytes2Base64 = Bijection$.MODULE$.bytes2Base64();
```
martijnhoekstra: Todo: To fix

bijection-core/src/test/scala/com/twitter/bijection/TypeclassBijectionLaws.scala
```diff
@@ -23,23 +23,6 @@ class TypeclassBijectionLaws
...
-          def negate(x: B) = bij(tc.negate(bij.invert(x)))
-          def fromInt(x: Int) = bij(tc.fromInt(x))
-          def toInt(x: B) = tc.toInt(bij.invert(x))
-          def toLong(x: B) = tc.toLong(bij.invert(x))
-          def toFloat(x: B) = tc.toFloat(bij.invert(x))
-          def toDouble(x: B) = tc.toDouble(bij.invert(x))
-          def compare(x: B, y: B) = tc.compare(bij.invert(x), bij.invert(y))
-        }
-    }
-
```
martijnhoekstra, line 42: Extending numeric requires an extra member in 2.13. Probably not worth it to split this test up, as typeclass testing seems adequately tested in other tests. Does it need additional replacements?

bijection-core/src/test/java/com/twitter/bijection/TestBijectionInJava.java
```diff
@@ -49,16 +49,18 @@ public void testBasicInjection() {
...
 
     @Test
     public void testBase64BijectionGzip() {
-        // Note, value classes return the underlying types in Java. But Java users usually
-        // don't care much about type safety, so punting on this for now
-        Bijection<byte[], String> bytes2GZippedBase64 = Bijection$.MODULE$.bytes2GZippedBase64();
+        //https://github.com/scala/scala/pull/8127
+        //generic AnyVal signatures pre-2.12.9 incorrectly had the underlying
+        //type, while in reality it's the boxed AnyVal
+        //Bijection<byte[], GZippedBase64String> bytes2GZippedBase64 = Bijection$.MODULE$.bytes2GZippedBase64();
```
martijnhoekstra: Todo: To fix

build.sbt
martijnhoekstra: I consolidated on 19.10.0

build.sbt
```diff
@@ -143,9 +185,12 @@ val ignoredABIProblems = {
     ),
     exclude[ReversedMissingMethodProblem](
       "com.twitter.bijection.twitter_util.UtilBijections.twitter2JavaFutureInjection"
-    )
-  )
-}
+    ),
+    exclude[DirectMissingMethodProblem]("com.twitter.bijection.netty.ChannelBufferBijection.invert"),
+    exclude[DirectMissingMethodProblem]("com.twitter.bijection.Bijection.trav2Vector")
```
martijnhoekstra: I don't understand why this is needed.

bijection-clojure/codegen/GeneratorWorkaroundJava.scala
```diff
@@ -0,0 +1,95 @@
+// This is a generator for java methods to forward to, as a workaround for
+// scala bug#11770
+//
+// Run this generator like a script:
+// scala GeneratorWorkaroundJava.scala > ../src/main/java/com/twitter/bijection/clojure/Workaround11770.java
```
martijnhoekstra: Todo: link to version-specific dir (other generators too)

build.sbt
```diff
@@ -128,6 +133,43 @@ def youngestForwardCompatible(subProj: String) = {
   }.toSet
 }
 
+/**
+ * Generic AnyVal's were incorrectly reported as the underlying type in
```
martijnhoekstra: @regadas these are the MiMa suppressions that work for the AnyVal bug from Java, if you want to do this independently first.

bijection-json/src/main/scala/com/twitter/bijection/json/JsonInjection.scala
```diff
@@ -73,7 +71,7 @@ trait LowPriorityJson {
   * You need to import all the methods of this object to get general
   * Injection[T,JsonNode] to work
   */
-object JsonNodeInjection extends LowPriorityJson with java.io.Serializable {
+object JsonNodeInjection extends CollectionJson with LowPriorityJson with java.io.Serializable {
```
martijnhoekstra, line 74: This approach, with version specific parts mixed in to a version-neutral "main" can also be applied to the other split files. The upside is less duplicated code. The downside is more complexity. Both are backward binary compatible.

Happy to do it either way.

martijnhoekstra:
Rebased to fix conflicts in tests

build.sbt
regadas: Hi @martijnhoekstra sorry for the late reply; Yeah I think we should do the `2.12.10` bump separate and add these rules.

martijnhoekstra: Sure, I'll make a separate PR

martijnhoekstra: @regadas https://github.com/twitter/bijection/pull/331

bijection-json/src/main/scala/com/twitter/bijection/json/JsonInjection.scala
martijnhoekstra: https://github.com/twitter/bijection/compare/develop...martijnhoekstra:213_unsplit?expand=1 is the "keep as much as possible together" approach

martijnhoekstra:
rebased

chrisbenincasa:
Any update on the status of this? This is blocking the conversion of the memcached module over in #771 (which seems to be the last dependencyless module)

martijnhoekstra:
If I remember right this is ready for review

dispalt:
This looks good anything keeping this from merging?

nevillelyh:
LGTM and MiMa is green too. @johnynek wanna take a final look?

johnynek:
This is great! Thanks for all the hard work!

martijnhoekstra:
Nice! Is a release also possible?

johnynek:
Just released:

https://github.com/twitter/bijection/releases/tag/v0.9.7

PR #288, opened 2019-08-23
build on openjdk8

martijnhoekstra:
Problem: Travis build validation doesn't work anymore because oracleJDK8 isn't supported anymore

Solution: Migrate to openjdk8

Effect: Working PR validation


johnynek:
Thank you!

PR #289, opened 2019-08-23
Upgrade to sbt1

martijnhoekstra:
Problem:

sbt 0.13 adoption and support is falling off, with many plugins dropping support, and scala 2.10 not being widely supported anymore

Solution:

Upgrade to sbt1


johnynek:
I'm willing to drop 2.10. If you are happy on 2.10, I guess you are happy with an old version of this library.

martijnhoekstra:
I had to do a double take on whether I actually hit comment on my remarks that I had no idea how aggregation exclusion wasn't working for 2.10, but I suppose you read my mind. I'll work on that over the weekend

johnynek:
Thanks for your help!

nevillelyh:
Can we get this merged? bijection is a dependency for chill & algebird.

johnynek:
Sorry for the latency.

We should probably add mima for 2.11 since it has different behavior than 2.12 but I’ll merge this and we can fix that later.

PR #326, opened 2019-10-14
Update scala versions 2.11.12 and 2.12.8

regadas:
We can't go up to `2.12.10`; there's a bug around value classes https://github.com/scala/bug/issues/11771 that will affect `Bijection` type when used in java.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -144,6 +144,12 @@ val ignoredABIProblems = {
     ),
     exclude[ReversedMissingMethodProblem](
       "com.twitter.bijection.twitter_util.UtilBijections.twitter2JavaFutureInjection"
+    ),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -143,6 +144,12 @@ val ignoredABIProblems = {
     ),
     exclude[ReversedMissingMethodProblem](
       "com.twitter.bijection.twitter_util.UtilBijections.twitter2JavaFutureInjection"
+    ),
+    exclude[DirectMissingMethodProblem](
```
regadas, line 148: MiMa false positives related to https://github.com/lightbend/mima/issues/388

johnynek (approved):

PR #331, opened 2019-10-16
bijection: 2.12.8 -> 2.12.10

martijnhoekstra:
Consequences for Java re-users:

2.12.9+ fixes a bug where `Bijection<A, B>` where `B` (or `A` for that matter) was a value class (`extends AnyVal`) was in the class file signature erroneously reported as the underlying type.

In reality, the Bijection returned an actual `AnyVal` instantiation.  Java code had to work around this bug by either casting the bijection, or casting the result.

This PR fixes the problem, which may break the java code dealing with the problem.

Binary compatibility is unaffected.

### DIFF
--- a/bijection-core/src/test/java/com/twitter/bijection/TestBijectionInJava.java
+++ b/bijection-core/src/test/java/com/twitter/bijection/TestBijectionInJava.java
@@ -49,16 +49,20 @@ public class TestBijectionInJava extends JUnitSuite {
     //TODO include a cleaner way to get to the scala Bijections than Bijection$.MODULE$.
     @Test
     public void testBase64Bijection() {
-        // Note, value classes return the underlying types in Java. But Java users usually
-        // don't care much about type safety, so punting on this for now
-        Bijection<byte[], String> bytes2Base64 = Bijection$.MODULE$.bytes2Base64();
+        //https://github.com/scala/scala/pull/8127
+        //generic AnyVal signatures pre-2.12.9 incorrectly had the underlying
+        //type, while in reality it's the boxed AnyVal
+        //the cast is required only for 2.11.x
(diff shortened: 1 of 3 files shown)


regadas (approved):
LGTM 👍

bijection-core/src/test/java/com/twitter/bijection/TestBijectionInJava.java
```diff
@@ -49,16 +49,20 @@ public void testBasicInjection() {
...
     @Test
     public void testBase64Bijection() {
-        // Note, value classes return the underlying types in Java. But Java users usually
-        // don't care much about type safety, so punting on this for now
-        Bijection<byte[], String> bytes2Base64 = Bijection$.MODULE$.bytes2Base64();
+        //https://github.com/scala/scala/pull/8127
+        //generic AnyVal signatures pre-2.12.9 incorrectly had the underlying
+        //type, while in reality it's the boxed AnyVal
+        //the cast is required only for 2.11.x
+        Bijection<byte[], Base64String> bytes2Base64 = (Bijection<byte[], Base64String>)(Object)Bijection$.MODULE$.bytes2Base64();
```
regadas, line 56: @martijnhoekstra  I think it's ok to remove `(Bijection<byte[], GZippedBase64String>)(Object)`

martijnhoekstra: That would be nice, wouldn't it? Unfortunately, the fix isn't backported to 2.11 (and most likely won't be -- 2.11 is EOL), so this is still needed on 2.11.x

regadas: @martijnhoekstra ahh yes right, nevermind!

PR #419, opened 2020-08-06
allow building on JDK 14

SethTisue:
on JDK 14, javac no longer accepts "-source 1.6" or "-target 1.6".

in general, Scala OSS projects no longer attempt to support JDK 6 or 7; I assume these flags weren't being kept around on purpose

this came up in the Scala 2.13 community build, here:
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk14-integrate-community-build/1127/artifact/logs/bijection-build.log

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -23,8 +23,8 @@ def scroogeSerializer =
 val buildLevelSettings = Seq(
   organization := "com.twitter",
   crossScalaVersions := Seq("2.11.12", scalaVersion.value, "2.13.1"),
-  javacOptions ++= Seq("-source", "1.6", "-target", "1.6"),
-  javacOptions in doc := Seq("-source", "1.6", "-Xlint:deprecation", "-Xlint:unchecked"),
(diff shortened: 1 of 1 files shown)


SethTisue:
CI failure looks like a transient failure to download a dependency

regadas (commented):
Setting it to `1.8` seems sensible to me these days. However, this setting is here because Twitter Hadoop needed this, but it might not be the case anymore? maybe @johnynek has more insights?

SethTisue:
(if it turns out 1.6 support is still needed, I could conditionalize it so that doing development on JDK 14+ is possible, but the 1.6 flags are still passed on JDK 8/11)

nevillelyh:
Maybe some tweep can comment, @mosesn any clue?

johnynek:
assuming this is fine. If they care they should pay attention.

mosesn:
Sorry, haven't checked my github notifs recently.  Yeah, we don't care about jdk6 or 7, but thanks for thinking of us :).  If you wanted to get rid of jdk8 we would be in trouble.

PR #425, opened 2020-08-20
[bijection-util] Adding bijection for Java CompletableFuture

jessefeinman:
Problem: The current implementation for converting Java Futures to Twitter Futures is resource-intensive and doesn't special-case Java CompletableFutures which don't need timers or threadpools like a plain Java Future does

Solution: Java Completable Futures are a special case of Java Futures which provide an easy way to do a callback to complete a Twitter Future without the overhead of an extra thread pool or timer

Result: Java CompletableFutures can be easily be converted to and from Twitter Futures

### DIFF
--- a/bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
+++ b/bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
@@ -16,7 +16,8 @@ limitations under the License.
 
 package com.twitter.bijection.twitter_util
 
-import java.util.concurrent.{Future => JavaFuture}
+import java.util.concurrent.{Future => JavaFuture, CompletableFuture}
+import java.util.function.BiConsumer
 
 import com.twitter.bijection._
 import com.twitter.io.Buf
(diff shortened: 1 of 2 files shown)


bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -133,6 +134,35 @@ trait UtilBijections {
...
+  /**
+    * Bijection between java completable futures and twitter futures.
+    */
+  implicit def twitter2CompletableFutureBijection[A]
+      : Bijection[TwitterFuture[A], CompletableFuture[A]] = {
+    new AbstractBijection[TwitterFuture[A], CompletableFuture[A]] {
+      override def apply(f: TwitterFuture[A]): CompletableFuture[A] = {
+        val cf = new CompletableFuture[A]()
+        f.respond {
+          case Return(value)    => cf complete value
```
johnynek: minor nit, can we avoid infix (which is going away in scala 3). `cf.complete(value)`

jessefeinman: No problem, do you want me to update the other places in this file where infix is being used? I mainly used it here due to its use elsewhere in the file.

johnynek: I understand.

No need to change the previous code. We can run a scalafix or scalafmt or something to get other use cases, I just don't want to add new ones. I don't want to complicate the PR with other changes, but also don't want to dig in deeper on the old pattern.

regadas (approved):

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -133,6 +134,35 @@ trait UtilBijections {
     }
   }
 
+  /**
+    * Bijection between java completable futures and twitter futures.
+    */
+  implicit def twitter2CompletableFutureBijection[A]
+      : Bijection[TwitterFuture[A], CompletableFuture[A]] = {
+    new AbstractBijection[TwitterFuture[A], CompletableFuture[A]] {
+      override def apply(f: TwitterFuture[A]): CompletableFuture[A] = {
```
mosesn, line 143: nit: we have a slightly optimized CompletableFuture impl in Future

https://github.com/twitter/util/blob/develop/util-core/src/main/scala/com/twitter/util/Future.scala#L2159

so we can simplify this to `f.toCompletableFuture`

bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala
```diff
@@ -133,6 +134,35 @@ trait UtilBijections {
...
+        val cf = new CompletableFuture[A]()
+        f.respond {
+          case Return(value)    => cf.complete(value)
+          case Throw(exception) => cf.completeExceptionally(exception)
+        }
+        cf
+      }
+
+      override def invert(f: CompletableFuture[A]): TwitterFuture[A] = {
+        val p = new TwitterPromise[A]()
```
mosesn, line 153: it would be cool to handle interruptions?  not a huge deal.

mosesn (commented):
sorry, this review was a bit late, figured I'd add some suggestions if we end up wanting to revisit this.

PR #518, opened 2021-06-27
Update libthrift to 0.14.2

regadas:


johnynek:
Thanks for doing this!

I really wonder if some of these modules are paying their way...

PR #565, opened 2021-12-16
Set scalafmt runner dialect

regadas:


johnynek:
Thanks for taking this!

